PR #19, opened 2022-11-07
Build ffmpeg and yt-dlp instead of using precompiled binaries

fsobolev:
As requested in flathub/org.nickvision.tubeconverter#7


nlogozzo:
LGTM, thanks!

PR #21, opened 2022-11-08
Fix downloading in good quality

fsobolev:
Currently downloading in good quality is totally broken, because nothing is passed to the "--format" flag, which results in invalid yt-dlp command.

According to the [yt-dlp documentation](https://github.com/yt-dlp/yt-dlp#format-selection-examples), it's possible to choose maximum video resolution. In my opinion, 1080p videos on YouTube are common enough to consider it as "minimal best", and so I consider 720p as "good", that's why I chose it in the code. But if you disagree, change it to whatever you think is appropriate, it's up to you.

### DIFF
--- a/src/models/download.cpp
+++ b/src/models/download.cpp
@@ -88,7 +88,7 @@ bool Download::download(bool embedMetadata)
     std::string cmd{ "" };
 	if (getMediaFileType().isVideo())
 	{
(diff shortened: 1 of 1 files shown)


nlogozzo:
I agree with the changes you've made. Thanks!

PR #22, opened 2022-11-08
Replace LongMessageDialog with LogsDialog

fsobolev:
After changes in LongMessageDialog, text is justified in center, which looks great in Tagger's advanced search info dialog, but it's not that good for showing logs IMO. I present you LogsDialog. It's basically the same as LongMessageDialog, but here are some changes:
* Uses TextView instead of Label, allowing to select text;
* Text is justified left, also it's monospace;
* Scrolls both vertically and horizontally (no wrapping).

There is one issue: text can't be selected using mouse, because it starts window dragging, but it can be selected using either keyboard shortcuts or right-click menu, which is better than not allowing selection at all I think. I'm not sure how to fix it.

Screenshot:


nlogozzo:
Maybe you can add a pill-button in the bottom right or top right of the dialog that says "Select All"?

fsobolev:
I think it's better to add "Copy to clipboard" button then, since it's the only reason a user might want to select the text. I'll be working on it tomorrow.

nlogozzo:
Sounds good to me!

src/ui/controls/logsdialog.cpp
```diff
@@ -0,0 +1,17 @@
...
+
+using namespace NickvisionTubeConverter::UI::Controls;
+
+LogsDialog::LogsDialog(GtkWindow* parent, const std::string& title, const std::string& logs, const std::string& cancelText, const std::string& destructiveText, const std::string& suggestedText) : MessageDialog(parent, title, "", cancelText, destructiveText, suggestedText)
+{
+    //Logs
+    m_scrolledWindow = gtk_scrolled_window_new();
+    gtk_widget_set_size_request(m_scrolledWindow, 500, 500);
+    m_textView = gtk_text_view_new();
+    m_textBuffer = gtk_text_view_get_buffer(GTK_TEXT_VIEW(m_textView));
```
nlogozzo, line 11: Could you also try adding `gtk_style_context_add_class(gtk_widget_get_style_context(m_textView), "card");` and see how the text view looks rounded?

src/ui/controls/logsdialog.hpp
```diff
@@ -19,10 +19,11 @@ namespace NickvisionTubeConverter::UI::Controls
     	 * @param destructiveText The text of the destructive button
     	 * @param suggestedText The text of the suggested button
     	 */
-		LongMessageDialog(GtkWindow* parent, const std::string& title, const std::string& description, const std::string& cancelText, const std::string& destructiveText = "", const std::string& suggestedText = "");
+		LogsDialog(GtkWindow* parent, const std::string& title, const std::string& description, const std::string& cancelText, const std::string& destructiveText = "", const std::string& suggestedText = "");
```
nlogozzo, line 22: Change `const std::string& description` to `const std::string& logs` here just like you did in the .cpp file.
Also update the documentation comment to reflect this change.

nlogozzo (changes requested):

fsobolev:
I've made all changes you asked, but it's not finished yet. I'm not familliar with C++, so I have a problem. I don't know how to properly connect button signal to a function that requires an argument:
```
g_signal_connect(m_copyButton, "clicked", G_CALLBACK((void (*)(GtkButton*, GtkTextBuffer*, gpointer))[](GtkButton*, GtkTextBuffer* m_textBuffer, gpointer data) { reinterpret_cast<LogsDialog*>(data)->onCopyToClipboard({ m_textBuffer }); }), this);
```
It doesn't work, the app just crashes without any error when I click the button.

src/ui/controls/logsdialog.cpp
```diff
@@ -1,17 +1,36 @@
...
     gtk_text_buffer_set_text(m_textBuffer, logs.c_str(), -1);
     gtk_text_view_set_editable(GTK_TEXT_VIEW(m_textView), FALSE);
     gtk_text_view_set_monospace(GTK_TEXT_VIEW(m_textView), TRUE);
     gtk_scrolled_window_set_child(GTK_SCROLLED_WINDOW(m_scrolledWindow), m_textView);
-    adw_message_dialog_set_extra_child(ADW_MESSAGE_DIALOG(m_gobj), m_scrolledWindow);
+    m_copyButton = gtk_button_new();
+    gtk_button_set_label(GTK_BUTTON(m_copyButton), _("Copy to clipboard"));
+    gtk_widget_add_css_class(m_copyButton, "pill");
+    gtk_widget_set_halign(m_copyButton, GTK_ALIGN_END);
+    g_signal_connect(m_copyButton, "clicked", G_CALLBACK((void (*)(GtkButton*, GtkTextBuffer*, gpointer))[](GtkButton*, GtkTextBuffer* m_textBuffer, gpointer data) { reinterpret_cast<LogsDialog*>(data)->onCopyToClipboard({ m_textBuffer }); }), this);
```
nlogozzo, line 25: So connect C++ functions to GTK C is a little annoying. Change this line to
` g_signal_connect(m_copyButton, "clicked", G_CALLBACK((void (*)(GtkButton*, GtkTextBuffer*, gpointer))[](GtkButton*, GtkTextBuffer* m_textBuffer, gpointer data) { reinterpret_cast<LogsDialog*>(data)->onCopyToClipboard(); }), this);`
without passing the `m_textBuffer` pointer to the function. Since inside the function, you already have access to `m_textBuffer`. This would require the method header to be: `void onCopyToClipboard();` with no parameter. Then from inside the function you will be able to use `m_textBuffer` without issues.

nlogozzo (changes requested):

fsobolev:
Done! It works.

nlogozzo:
I also just extended the margins a little to make it look better:
image

PR #23, opened 2022-11-08
Added Arabic translate.

fawaz006:
I'm really didn't know how to build it and I also get error that say jsoncpp aren't found even it was installed, so I just coped fr.po and edit it, so I don't know if there's somthing wrong or all are ok

### DIFF
--- a/po/ar.po
+++ b/po/ar.po
@@ -223,7 +223,7 @@ msgstr "محول"
 
 #: src/ui/views/preferencesdialog.cpp:31
 msgid "Customize converter settings."
-msgstr "Customize converter settings."
+msgstr "تخصيص إعدادات المحول."
 
 #: src/ui/views/preferencesdialog.cpp:36
 msgid "Embed Metadata"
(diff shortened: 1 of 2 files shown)


po/ar.po
```diff
@@ -0,0 +1,246 @@
...
+msgid "Dark"
+msgstr "مظلم"
+
+#: src/ui/views/preferencesdialog.cpp:30
+msgid "Converter"
+msgstr "محول"
+
+#: src/ui/views/preferencesdialog.cpp:31
+msgid "Customize converter settings."
+msgstr "Customize converter settings."
```
nlogozzo, line 226: I think you forgot to translate here.

nlogozzo (changes requested):
Line 226

po/ar.po
fawaz006: wired, I remember that I translate it?

nlogozzo:
LGTM, thanks!

PR #24, opened 2022-11-09
Edit on download.cpp

fawaz006:
Meow

### DIFF
--- a/src/models/download.cpp
+++ b/src/models/download.cpp
@@ -109,7 +109,7 @@ bool Download::download(bool embedMetadata)
 	}
 	if(embedMetadata)
 	{
-	    cmd += " --add-metadata --embed-chapters --write-description --embed-thumbnail";
+	    cmd += " --add-metadata --embed-chapters --embed-thumbnail";
 	}
 	setLog("URL: " + getVideoUrl() + "\nPath: " + getSavePath() + "\nQuality: " + std::to_string(static_cast<int>(getQuality())) + "\n");
 	std::pair<int, std::string> result{ CmdHelpers::run(cmd, "r", m_pid) };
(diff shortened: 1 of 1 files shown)


fsobolev:
I agree that "--embed-chapters" is a good addition, but "--write-description"? It's usually not needed by most users I think. @nlogozzo, what about implementing this as an option?

nlogozzo:
What does the `write-description` flag do? Just writes the description of the video on the tag? I think that would be valid to have since it is metadata on the video. However sometimes descriptions are long and useless.

fawaz006:
> What does the write-description flag do? Just writes the description of the video on the tag?

Yeah, even it might not be very useful, but it is under the “Metadata” option anyway.

fsobolev:
> What does the `write-description` flag do? Just writes the description of the video on the tag? I think that would be valid to have since it is metadata on the video. However sometimes descriptions are long and useless.

No, not in the tag, it creates `*.description` text file. That's why I'm against it.

fawaz006:
> No, not in the tag, it creates *.description text file. That's why I'm against it.

I was thinking it's just writes the description of the video on the tag
it's look, there are no need to it👍

nlogozzo:
Oh if it creates an extra file then yeah we don't need it. Is there an --embed-description flag of some kind?

fsobolev:
I don't see anything like this in yt-dlp docs. But actually, it seems we don't need `--embed-chapters` since `--embed-metadata` already includes it:
```
--embed-metadata                Embed metadata to the video file. Also
                                embeds chapters/infojson if present unless
                                --no-embed-chapters/--no-embed-info-json are
                                used (Alias: --add-metadata)
```

nlogozzo:
Oh. So then this PR does nothing essentially?

fsobolev:
Sadly, yes 😢

fawaz006:
Yeah, I checked it and it's already embed
image

PR #48, opened 2023-02-10
Add snap-store link also

soumyaDghosh:


fsobolev:
> https://snapcraft.io/tube-converter

404 error.

Also, badges doesn't look good because of different height.

nlogozzo (approved):

PR #52, opened 2023-02-13
Save Path

fsobolev:
This PR merges "Save Folder" and "New Filename" rows into one - "Save Path".

The entry itself is not editable, a path can only be changed using file chooser. This way a path can't be empty: either it's set by a user, or the controller generates new path using previously used folder (which is now never empty too, it defaults to Videos folder), a video title and selected file format. File chooser is filtered to show only files with selected format. During validation, all rows except URL become insensitive.
Home access removed from flatpak.

### DIFF
--- a/NickvisionTubeConverter.GNOME/Views/AddDownloadDialog.cs
+++ b/NickvisionTubeConverter.GNOME/Views/AddDownloadDialog.cs
@@ -2,6 +2,7 @@ using NickvisionTubeConverter.Shared.Controllers;
 using NickvisionTubeConverter.Shared.Models;
 using System;
 using System.IO;
+using System.Runtime.InteropServices;
 using System.Threading.Tasks;
 
 namespace NickvisionTubeConverter.GNOME.Views;
@@ -9,8 +10,15 @@ namespace NickvisionTubeConverter.GNOME.Views;
 /// <summary>
 /// The AddDownloadDialog for the application
 /// </summary>
-public class AddDownloadDialog
+public partial class AddDownloadDialog
 {
+    [LibraryImport("libadwaita-1.so.0", StringMarshalling = StringMarshalling.Utf8)]
+    [return: MarshalAs(UnmanagedType.I1)]
+    public static partial bool gtk_file_chooser_set_current_folder(nint chooser, nint file, nint error);
+
+    [LibraryImport("libadwaita-1.so.0", StringMarshalling = StringMarshalling.Utf8)]
+    public static partial void gtk_file_chooser_set_current_name(nint chooser, string name);
+
     private bool _constructing;
     private readonly Gtk.Window _parent;
     private readonly AddDownloadDialogController _controller;
@@ -114,11 +122,11 @@ public class AddDownloadDialog
         _btnSelectSavePath.SetValign(Gtk.Align.Center);
         _btnSelectSavePath.AddCssClass("flat");
         _btnSelectSavePath.SetIconName("folder-open-symbolic");
-        _btnSelectSavePath.SetTooltipText(_controller.Localizer["SelectSaveFolder"]);
+        _btnSelectSavePath.SetTooltipText(_controller.Localizer["SelectSavePath"]);
         _btnSelectSavePath.OnClicked += SelectSavePath;
         _rowSavePath = Adw.EntryRow.New();
         _rowSavePath.SetSizeRequest(420, -1);
-        _rowSavePath.SetTitle(_controller.Localizer["SaveFolder", "Field"]);
+        _rowSavePath.SetTitle(_controller.Localizer["SavePath", "Field"]);
         _rowSavePath.AddSuffix(_btnSelectSavePath);
         _rowSavePath.SetEditable(false);
         _preferencesGroup.Add(_rowSavePath);
@@ -131,7 +139,7 @@ public class AddDownloadDialog
         if(string.IsNullOrEmpty(_rowSavePath.GetText()))
         {
             _rowSavePath.AddCssClass("error");
-            _rowSavePath.SetTitle(_controller.Localizer["SaveFolder", "Invalid"]);
+            _rowSavePath.SetTitle(_controller.Localizer["SavePath", "Invalid"]);
         }
         _dialog.SetResponseEnabled("ok", false);
         _constructing = false;
@@ -170,7 +178,7 @@ public class AddDownloadDialog
         _rowVideoUrl.RemoveCssClass("error");
         _rowVideoUrl.SetTitle(_controller.Localizer["VideoUrl", "Field"]);
         _rowSavePath.RemoveCssClass("error");
-        _rowSavePath.SetTitle(_controller.Localizer["SaveFolder", "Field"]);
+        _rowSavePath.SetTitle(_controller.Localizer["SavePath", "Field"]);
         _rowSavePath.SetText(_controller.GetSavePath());
         if (checkStatus == DownloadCheckStatus.Valid)
         {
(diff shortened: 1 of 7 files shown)


NickvisionTubeConverter.Shared/Models/Configuration.cs
```diff
@@ -49,7 +49,7 @@ public Configuration()
             Directory.CreateDirectory(ConfigDir);
         }
         Theme = Theme.System;
-        PreviousSaveFolder = "";
+        PreviousSaveFolder = Environment.GetFolderPath(Environment.SpecialFolder.MyVideos);
```
nlogozzo, line 52: I don't think it is the best choice to default to `MyVideos`. What is users use `Tube Converter` for downloading videos as audio files...`MyVideos` wouldn't make sense here. Best to leave it blank and have it set when they first chose the folder, no?

nlogozzo:
Thanks for this. Haven't tested the functionality or anything yet...will do as soon as i get a chance :)

NickvisionTubeConverter.Shared/Models/Configuration.cs
fsobolev: This value is used only once - when a user opens add download dialog for the first time, then it gets replaced by whatever folder they choose after starting a download. And `Videos` seems to be sane default, because probably most users will want to download a video, not audio. Setting this to `home` or anything else "neutral" is even worse, because it will not be appropriate for most cases, I think that we might expect with high chances that users want to save files in eiher `Videos` or `Music` directory or somewhere on external drive, and the first one, while not being perfect for all use-cases, is the best default.
Setting it to blank doesn't solve anything but creates 2 problems:
1. All users will be forced to select a folder on the first download, so it doesn't help anybody, but `Videos` folder can be suitable for a lot of users;
2. Controller now expects `PreviousSaveFolder` to never be empty, because when it sees that URL has changed it generates new path with video title as filename. If there are no folder set, it can't generate a correct path.

nlogozzo: Ah since the controller expects a value now then no worries we can leave it as is :)

DaPigGuy:
Using the Save file chooser action, you now have to explicitly grant permission for every file path to save, which is not at all implied by the Add Download dialog. Might it be better to forgo the Save Path entry in the dialog and simply open the file chooser on clicking Ok?

fsobolev:
Ah, I think I see the problem now... Idk, maybe it wasn't there because I had home permission on Tube Converter installed on the system, but after I disabled it I see portal's path instead of actual save path in the row

fsobolev:
Either we keep home permission to make saving previous folder work properly inside home (but not outside, because the app will remember portal's path then)... Or we remake saving on a manner of web browsers. Instead of remembering last used folder, the app will remember a folder that a user selected in settings. There will be no save path row in the dialog, but instead after pressing "OK" (reminder: we need to rename it to follow the GNOME HIG) a file chooser will appear with the selected folder opened and with video title as filename. We could also add "autosave" toggle to not show file chooser and save files automatically.
What do you think?

nlogozzo:
I don't see a problem with keeping `home` permissions...

fsobolev:
> I don't see a problem with keeping `home` permissions...

I think instead we can give access to `Music`, `Video` and `Downloads`, and if a user tries to save a file somewhere else then the path shouldn't be remembered and a warning button should appear in the row to notify a user that the path will not be saved and they need to use Flatseal if they want to grant the app a permission to remember this path.

nlogozzo:
> > I don't see a problem with keeping `home` permissions...
> 
> I think instead we can give access to `Music`, `Video` and `Downloads`, and if a user tries to save a file somewhere else then the path shouldn't be remembered and a warning button should appear in the row to notify a user that the path will not be saved and they need to use Flatseal if they want to grant the app a permission to remember this path.

Sounds like the best option to me!
But this should only be used for when the app is running on flatpak not via AUR or manually ofc...

fsobolev:
> But this should only be used for when the app is running on flatpak not via AUR or manually ofc...

If the requested path is not permitted and portal is used, then the path will start with `/run/user/$UID/doc/`, this will not be the case for non-flatpak builds.
Idk how to handle it in snap tho.

fsobolev:
[Kooha-2023-02-14-06-02-44.webm](https://user-images.githubusercontent.com/117388856/218628535-c6f71486-f061-4e82-8ab4-9380a1adef88.webm)

Maybe wording for the warning could be better, but everything else seems to be finished on the controller and GNOME side. Only WinUI is left.

NickvisionTubeConverter.Shared/Resources/Strings.resx
```diff
@@ -369,4 +369,10 @@
 	<data name="Summary.GTK" xml:space="preserve">
 		<value>Download your favorite videos</value>
 	</data>
+
+	<data name="SaveWarning.GTK" xml:space="preserve">
+		<value>Looks like you want to save a file in a folder outside of Flatpak sandbox.
```
nlogozzo, line 374: Better wording:
```
The selected save folder is outside the Flatpak sandbox. The file will be saved, but the folder path will not be remembered for future use&#xA;If you want to always remember this folder, give Tube Converter flatpak write permissions to the selected folder. (An application such as Flatseal can be used to achieve this task)
```

nlogozzo (changes requested):

nlogozzo:
image

The subtitle row is not being disabled when the file type is audio.

NickvisionTubeConverter.GNOME/Views/AddDownloadDialog.cs
```diff
@@ -173,8 +173,16 @@ public AddDownloadDialog(AddDownloadDialogController controller, Gtk.Window pare
...
+        _rowFileType.SetSensitive(false);
+        _rowSavePath.SetSensitive(false);
+        _rowQuality.SetSensitive(false);
+        _rowSubtitle.SetSensitive(false);
         var checkStatus = await _controller.UpdateDownloadAsync(_rowVideoUrl.GetText(), (MediaFileType)_rowFileType.GetSelected(), _rowSavePath.GetText(), (Quality)_rowQuality.GetSelected(), (Subtitle)_rowSubtitle.GetSelected());
         _spinnerVideoUrl.Stop();
+        _rowFileType.SetSensitive(true);
+        _rowSavePath.SetSensitive(true);
+        _rowQuality.SetSensitive(true);
+        _rowSubtitle.SetSensitive(true);
```
nlogozzo, line 212: This should be
```csharp
_rowSubtitle.SetSensitive(((MediaFileType)_rowFileType.GetSelected()).GetIsVideo());
```

nlogozzo: Line 82 can then be removed as well.

nlogozzo (changes requested):

nlogozzo:
Also after making those changes, if you could sync this branch with the latest commits in `csharp`...i'll merge in the morning :)

For now, i must sleep 🫡

NickvisionTubeConverter.GNOME/Views/AddDownloadDialog.cs
fsobolev: Line 82 is
```
_rowFileType.SetModel(Gtk.StringList.New(new string[] { "MP4", "WEBM", "MP3", "OPUS", "FLAC", "WAV" }));
```

why remove?

nlogozzo: Oh sorry I got the nunber wrong. Is should be whatever the line is that duplicates that same line I provided in the OnNotify event handler of rowSubtitle.

nlogozzo (approved):

PR #50, opened 2023-02-11
C++ --> C#/.NET

nlogozzo:
Ports the C++ version as is to C# as well as adding a WinUI version and fixes minor issues.

Closes #49 
Closes #42 
Closes #41 
Closes #40 
Closes #13 

---
# TODO
## Shared
- [X] `Download` model implementation
- [X] `MainWindowController` implementation
- [X] Use `pythonnet` for `yt-dlp` (#54)
## GNOME
- [X] `Add Download` dialog submit
- [x] `Downloads` page
  - [x] Download entry
    - [x] Downloading
- [X] Prevent closing during download
- [x] Retry Button
  - [x] Icon
- [x] Adaptive
- [X] New Sun/Moon Icon
## WinUI
- [X] Everything 😆

PR #53, opened 2023-02-17
Create spanish translation

bruno-cabrera98:


DaPigGuy:
I believe we are transitioning to Weblate after the transition to C# (#50) which should be getting merged soon, so the translations will have to wait until then. Thanks for the contributions though :slightly_smiling_face: 

@nlogozzo

nlogozzo:
Yes, sorry to make you go through that translation for nothing...but Tube Convertrr C# Rewrite should be merged by Sunday and will require weblate translation :)

I'll ping here once it's ready!

PR #54, opened 2023-02-20
Rewritten Backend In Python.NET

nlogozzo:
This allows the backend to work for WinUI as well :)


NickvisionTubeConverter.GNOME/Controls/DownloadRow.cs
```diff
@@ -158,33 +158,31 @@ public async Task StartAsync(bool embedMetadata)
         _imgStatus.SetFromIconName("folder-download-symbolic");
         _viewStackState.SetVisibleChildName("downloading");
         _progLabel.SetText(_localizer["DownloadState", "Preparing"]);
-        var success = await _download.RunAsync(embedMetadata, new Progress<DownloadProgress>((x) =>
+        _lblFilename.SetText(_download.Filename);
```
DaPigGuy, line 161:
```suggestion
```

No longer needed, was a remnant from before we fetched the video title for the default file name before creating the row

NickvisionTubeConverter.Shared/Models/Download.cs
```diff
@@ -123,91 +124,99 @@ public static async Task<string> GetVideoTitleAsync(string url)
...
+            var result = new DownloadProgressState()
+            {
+                Status = entries["status"].As<string>() switch
+                {
+                    "processing" => ProgressStatus.Processing,
+                    "downloading" => ProgressStatus.Downloading,
+                    _ => ProgressStatus.Other
+                },
+                Progress = entries.HasKey("downloaded_bytes") ? entries["downloaded_bytes"].As<double>() / entries["total_bytes"].As<double>() : 0,
+                Speed = entries.HasKey("speed") ? entries["speed"].As<double>() : 0
```
fsobolev, line 214: if during download the speed is unknown (in case of connection problems for example), the key `speed` will exist, but `entries["speed"]` will be `None` (python's equivalent to null), that can't be converted to double, this will cause crash. Idk if `As<double?>()` will work, that's why I used try-catch.

fsobolev (changes requested):

NickvisionTubeConverter.Shared/Models/Download.cs
fsobolev: Works good, I tested it by manually setting `entries["speed"]` to `None`

nlogozzo:
Not ready yet sorry...forgot to fix stopping downloads

NickvisionTubeConverter.Shared/Helpers/DependencyManager.cs
```diff
@@ -68,19 +42,37 @@ public static string Ffmpeg
...
-        {
-            return true;
-        }
         try
         {
-            await YoutubeDL.DownloadYtDlpBinary(YtdlpPath.Remove(YtdlpPath.IndexOf("yt-dlp.exe")));
-            await YoutubeDL.DownloadFFmpegBinary(Ffmpeg.Remove(Ffmpeg.IndexOf("ffmpeg.exe")));
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                await YoutubeDL.DownloadFFmpegBinary(Ffmpeg.Remove(Ffmpeg.IndexOf("ffmpeg.exe")));
```
DaPigGuy, line 54: Probably shouldn't depend on YoutubeDLSharp solely to install FFmpeg :sweat_smile:

fsobolev (changes requested):
> The progress bar begins pulsing extremely quickly after cancelling &
retrying multiple times due to several timeouts all pulsing the
progress bar.

It's probably because cancellation doesn't work 😅 When we'll fix it the progress bar should be fine.
Old implementation (that you returned) doesn't work properly, for some reason `_progBar.Pulse()` only takes effect on the main callback call, not on every `g_timeout_add` tick.

DaPigGuy:
>  for some reason _progBar.Pulse() only takes effect on the main callback call, not on every g_timeout_add tick

Huh, can't believe I didn't notice that before. Looks like the status keeps switching between `Processing` and `Other`

fsobolev:
> Looks like the status keeps switching between `Processing` and `Other`

I tried to call `pulse` for both statuses, it didn't help. Some weird magical threading issue is here I guess.
But as I said, as soon as we manage to make cancelling work everything should be fine.

NickvisionTubeConverter.Shared/Helpers/DependencyManager.cs
nlogozzo: Yes I will fix that before merge :)

DaPigGuy:
Setting the yt-dlp `finished` status as our own `Processing` status seems to work fine on my end.

fsobolev:
> Setting the yt-dlp finished status as our own Processing status seems to work fine on my end.

No, still broken. Well, it works better than without "finished", but it moves abruptly.

DaPigGuy:
Well, that was embarrassingly not thorough testing on my part... This latest commit should solve all our progress bar pulsing problems :sweat_smile:

DaPigGuy:
Latest change cuts about 1 second from the time for the OK button to become sensitive :upside_down_face:

nlogozzo:
> Latest change cuts about 1 second from the time for the OK button to become sensitive (for me) :upside_down_face:

Hey, every second counts!

I just got home from work going to try and get stop downloads working :)
And replace youtubedlsharp for ffmpeg download

NickvisionTubeConverter.Shared/Models/Download.cs
```diff
@@ -97,49 +97,18 @@ public static async Task<bool> GetIsValidVideoUrlAsync(string url)
                         { "quiet", true },
                         { "merge_output_format", "/" }
                     };
-                    ytdlp.YoutubeDL(ytOpt).extract_info(url, download: false);
+                    Python.Runtime.PyDict videoInfo = ytdlp.YoutubeDL(ytOpt).extract_info(url, download: false);
+                    return (true, videoInfo.HasKey("title") ? (videoInfo["title"].As<string>() ?? "Video") : "Video");
```
fsobolev, line 101: `As<string?>()` will make it become null if it's None in python, so it will not raise null exception and the block will return "Video" as expected.

fsobolev (changes requested):

NickvisionTubeConverter.Shared/Models/Download.cs
nlogozzo: Yes the `??` operator means if this object is null use this value instead so in this case we want, if the string is null (None in python) return "Video".

Using `string` for `As<>` instead of `string?` will never cause it actually be null, it will just crash as @fsobolev mentioned.

nlogozzo:
Okay...if all looks well, we can merge :)

NickvisionTubeConverter.WinUI/Views/MainWindow.xaml.cs
```diff
@@ -212,6 +231,6 @@ private void NotificationSent(object? sender, NotificationSentEventArgs e)
     /// <param name="e">RoutedEventArgs</param>
     private async void AddDownload(object sender, RoutedEventArgs e)
     {
-        
+
```
fsobolev, line 234: empty function?

nlogozzo: That will be for the AddDownloadDialog when I implement it in WinUI...but that's something for #50 not this PR :)

fsobolev (approved):

PR #58, opened 2023-02-26
Blueprints & Just

DaPigGuy:
Let's adopt [Blueprints](https://jwestman.pages.gitlab.gnome.org/blueprint-compiler/index.html)! Far more organized and cleaner UI declaration than the previous mess...

**TODO**
- [x] Main Window
- [x] Add Download Dialog
- [x] Download Row
- [x] Preferences Dialog
- [x] Shortcuts Dialog
- [x] Add Just Buildsystem

**Notes**
I won't be porting NickvisionTubeConverter.GNOME.Controls.MessageDialog for the time being, I'd like to get Adw.MessageDialog working in place of it for the Quit prompt later.


NickvisionTubeConverter.GNOME/Views/Builder.cs
```diff
@@ -0,0 +1,37 @@
...
+                {
+                    element.RemoveAttribute("translatable");
+                    element.InnerText = localizer[element.InnerText];
+                }
+            }
+            builder = Gtk.Builder.NewFromString(xml.OuterXml, -1);
+        }
+        return builder;
+    }
+}
```
fsobolev, line 37: I'm not sure if this belongs to `Views`. I would create `Helpers` namespace and directory for this, similar to what we have in `Shared`. @nlogozzo what do you think?

NickvisionTubeConverter.GNOME/Views/MainWindow.cs
```diff
@@ -50,162 +35,57 @@ public MainWindow(MainWindowController controller, Adw.Application application)
...
-        _binSpinner.SetChild(_spinner);
-        //Loading Overlay
-        _overlayLoading = Gtk.Overlay.New();
-        _overlayLoading.SetVexpand(true);
-        _overlayLoading.AddOverlay(_binSpinner);
-        _overlayLoading.SetChild(_mainBox);
-        Handle.SetContent(_overlayLoading);
+        OnCloseRequest += OnCloseRequested;
+        //Build UI
+        Gtk.Builder builder = Builder.FromFile("window.ui", _controller.Localizer);
```
fsobolev, line 47: is explicit type required? The method can only return `Gtk.Builder`, so `var` here should work I think

fsobolev (changes requested):

NickvisionTubeConverter.GNOME/Views/Builder.cs
nlogozzo: I was just going to say the same,
I would make a folder in `NickvisionTubeConverter.GNOME` called `Helpers` and move this class there.
*Make sure to change the namespace when you move the file as well :) *

nlogozzo (approved):

fsobolev (approved):

fsobolev (approved):

DaPigGuy:
Looks like the Blueprint compiler is no longer being run?

fsobolev:
> Looks like the Blueprint compiler is no longer being run?

Oh god, sorry, I'm sleepy, forgot about it!

fsobolev:
@DaPigGuy

nlogozzo (approved):

PR #62, opened 2023-02-27
Queue System

nlogozzo:
This PR implements a queue system for downloads similar to that seen in Fragments.
This is the preliminary work needed before we implement playlist support.

Downloads can be placed in 3 sections and will be displayed in this order:
- Downloading 
- Completed
- Queue

There will be a new setting called `Max Number of Active Downloads` in which the user can choose between 1-10 (5 being the default) of the number of downloads to be run at a time. 
If there are the max number of active downloads running in the downloading section, the download will be placed in the queue and will wait until an active download finishes to then be moved to the downloading section.

image
image


nlogozzo:
I would like to release beta2 after this is merged to allow people to test the new queue system. And then work on playlist support :)

nlogozzo:
https://github.com/nlogozzo/NickvisionTubeConverter/releases/tag/2023.3.0-beta2

PR #82, opened 2023-03-11
Playlist and Shell Notification Support

### LINKED ISSUE #2: Add Playlist Download Feature

### LINKED ISSUE #73: Show Notification When Window Inactive And Downloads Finish

nlogozzo:
This PR implements support for downloading playlists.

Closes #2 
Closes #73 

## TODO
- [x] Backend Implementation
- [x] GNOME Implementation
- [x] WinUI Implementation
- [x] Handle overwriting files
- [x] Button to add numbers to titles of download
- [x] Shell notification support


nlogozzo:
@DaPigGuy thank you for the commit...you just made my life 100x easier 😅

NickvisionTubeConverter.Shared/Models/VideoUrlInfo.cs
```diff
@@ -61,14 +66,26 @@ private VideoUrlInfo(string url)
...
                     {
-                        videoUrlInfo.IsPlaylist = true;
+                        var entries = videoInfo["entries"].As<Python.Runtime.PyList>();
+                        foreach (var e in entries)
+                        {
+                            var entry = e.As<Python.Runtime.PyDict>();
+                            videoUrlInfo.Videos.Add(new VideoInfo()
+                            {
+                                Url = entry["webpage_url"].As<string>(),
+                                Title = entry.HasKey("title") ? entry["title"].As<string>() ?? "Video" : "Video"
```
fsobolev, line 90: I don't mind that we are "youtube-oriented" in the first place when thinking about TC and writing the code, but I think we should also keep in mind that yt-dlp is capable of downloading audio files (I'm not talking about extracting audio from video, but about downloading media files that are audio from the beginning - files from bandcamp, for example). And so the default title should be neutral in that regard.
This problem also applies to the app's description.

DaPigGuy: Perhaps, `Unknown` or `Media`?

nlogozzo: > I'm not talking about extracting audio from video, but about downloading media files that are audio from the beginning - files from bandcamp, for example

We should probably add a check for this and disable selecting video formats as well if this is the case.

> Perhaps, Unknown or Media?

I think `Media` would be good.

> This problem also applies to the app's description.

I mean maybe the description is ok since youtube-dl's description is: "Command-line program to download videos from YouTube.com and other video sites"

fsobolev: I like `Media` for single file. For playlist, maybe `Playlist Item` + numbering feature suggested in the issue should be force-enabled, otherwise only one file will be downloaded.

nlogozzo:
Here is playlist support. Implemented in backend and WinUI. All that's left to do is GNOME. Here you can see first downloading a playlist, selecting certain videos and changing the titles of the downloads, as well as then downloading just a single video.

https://imgur.com/a/Ko26kWX

NickvisionTubeConverter.GNOME/Views/AddDownloadDialog.cs
```diff
@@ -41,39 +41,183 @@ public partial class AddDownloadDialog : Adw.MessageDialog
     private delegate void GAsyncReadyCallback(nint source, nint res, nint user_data);
 
     [LibraryImport("libadwaita-1.so.0", StringMarshalling = StringMarshalling.Utf8)]
-    private static partial void gtk_file_dialog_save(nint dialog, nint parent, nint cancellable, GAsyncReadyCallback callback, nint user_data);
+    private static partial void gtk_file_dialog_select_folder(nint dialog, nint parent, nint cancellable, GAsyncReadyCallback callback, nint user_data);
 
     [LibraryImport("libadwaita-1.so.0", StringMarshalling = StringMarshalling.Utf8)]
-    private static partial nint gtk_file_dialog_save_finish(nint dialog, nint result, nint error);
+    private static partial nint gtk_file_dialog_select_folder_finish(nint dialog, nint result, nint error);
```
fsobolev, line 47: We still need `save` for selecting a path for single media as was discussed in matrix.

fsobolev (changes requested):

NickvisionTubeConverter.GNOME/Views/AddDownloadDialog.cs
nlogozzo: The single media works the same way as a playlist just with one download instead of 5, similar to WinUI if you look at that. So there is no need for `save`.

fsobolev: Well, I thought will continue to use "Save path" for single media and use "Save folder" only for playlists. But if you want to revert it, okay. No checks of file existence again tho.

nlogozzo: The problem is it would be too complicated in the backend to separate `save path` vs `save folder` when dealing with single video vs playlist. We could add another row to the AddDownloadDialog that's a simple toggle called `Overwrite Existing Files` which we can pass to yt-dlp per download whether to overwrite files or not.

Because remember, if i'm downloading an `mp3` there will first be an `mp4` downloaded and then a `mp3`. So with the `save` dialog if I was to agree to overwriting the `mp3` file, the `mp4` file will be implicitly overwritten tho which the user might not want.

So i think this toggle might be the best solution...

fsobolev: From yt-dlp:
```
paths:             Dictionary of output paths. The allowed keys are 'home'
                       'temp' and the keys of OUTTMPL_TYPES (in utils.py)
```
So we can have all temporary files in separate folder and not worry about overriding in cases such as you provided. Then only the target file overriding should be considered, and so we don't need a toggle, only checking with `File.Exists(folder+name+ext)` and showing a warning if the folder is opened through portal.

nlogozzo: so if said file exists, we will have a label saying something like `There are files in the save folder that conflict with the titles of some videos to be downloaded. These files will be overwritten. To prevent this, change the title of the video to download.`

fsobolev: > so if said file exists, we will have a label saying something like `There are files in the save folder that conflict with the titles of some videos to be downloaded. These files will be overwritten. To prevent this, change the title of the video to download.`

No. We should set `temp` in `paths` dict to something like `/tmp/Nickvision/TubeConverter` (using available dotnet variable to get the equivalent of `/tmp` on Windows) so all files except target file will be placed there.
Then for target file only we should produce 2 checks:
1. `File.Exists()`
2. If a folder is not available for flatpak and so accessible though portal. We already have this check to show a warning that the folder will not be remembered, and this warning should be extended to provide information that it's impossible to check if the file exists (because `File.Exists()` will always return `false` in portalled folder, since the actual content of the real folder can't be read).

nlogozzo: > We should set temp in paths dict to something like /tmp/Nickvision/TubeConverter (using available dotnet variable to get the equivalent of /tmp on Windows

We have the config path where we could create a temp folder there and use, to be consistent across platforms.

nlogozzo: I committed an implementation of the overwrite toggle in the add download dialog. I feel like it would be a lot easier from a user's perspective to have a toggle switched on and off depending on whether or not they will want to overwrite files and pass that logic to yt-dlp, instead of going through flatpak permissions and warning the user with all that complicated sandbox stuff. According to yt-dlp, if overwrite is false it won't override both the mp4 and mp3 files used for downloading a mp3 file, as described in the previous scenario.

Check it out and let me know what you think, hope we can find common ground on that :)

fsobolev:
This time I actually run the app instead of just looking at the code.
1. I wanted to download a single video, but it was detected as playlist. Probably `entries` key exists even for single video? We should revert to `playlist-count` for detection then.
2. I don't like how numbering is implemented. What if a user changes their mind and wants to disable it? I think this should be a toggle button that adds `N - ` to action rows, but not to entries in popovers. This will require to not use properties bindings but instead do something like this:
    ```
    _numberedRegex = new Regex(@"[0-9]? - ", RegexOptions.None);
    ...
    entry.OnNotify += (sender, e) =>
    {
        if (e.Props.GetName() == "text")
        {
            row.SetTitle((_numbered ? _numberedRegex.Match(row.GetTitle()) : "") + entry.GetText());
        }
    };
    ```
    This way a user will not be able to break numbering.
3. I agree with your logic on adding overwrite toggle, but I think we still should add `temp` path - this way when overwriting is disabled a user will not get his mp4 file overwritten unexpectedly because they decided to download mp3 version of the same file.

nlogozzo:
> I wanted to download a single video, but it was detected as playlist. Probably entries key exists even for single video? We should revert to playlist-count for detection then.

Can I have the url of this video?

> I don't like how numbering is implemented. What if a user changes their mind and wants to disable it? I think this should be a toggle button that adds N -  to action rows, but not to entries in popovers. This will require to not use properties bindings but instead do something like **this:**

agreed 👍

> this way when overwriting is disabled a user will not get his mp4 file overwritten unexpectedly because they decided to download mp3 version of the same file.

As I said, according to the [yt-dlp docs](https://github.com/yt-dlp/yt-dlp/blob/master/yt_dlp/YoutubeDL.py#L246) and my understanding, that mp4 file will not be overwritten if it's set to false, even if downloading an mp3. Will double check

> Up to you, but I would get rid of empty space when the dialog is first opened. To acieve this, all children of the stack should be hidden at start.

Yeah, I didn't know how to fix this...but thanks to you, now I do :)

fsobolev:
> Can I have the url of this video?

https://www.youtube.com/watch?v=V3lbZBilfJY

> As I said, according to the [yt-dlp docs](https://github.com/yt-dlp/yt-dlp/blob/master/yt_dlp/YoutubeDL.py#L246) and my understanding, that mp4 file will not be overwritten if it's set to false, even if downloading an mp3. Will double check

My bad, I meant if overwriting is enabled. A user can not realize that for some services (not youtube) extracting audio will require downloading video first, and so they will not disable overwriting,

nlogozzo:
> My bad, I meant if overwriting is enabled. A user can not realize that for some services (not youtube) extracting audio will require downloading video first, and so they will not disable overwriting,

Ah makes sense, ok i'll do that.

> https://github.com/nlogozzo/NickvisionTubeConverter/pull/82/commits/1a9794f1f2f5c7ed1b0e9820ab8f9409f99a6099

If you could test this commit on GNOME :) I'm not running the full DE so I can't test on WSL.

nlogozzo:
> Up to you, but I would get rid of empty space when the dialog is first opened. To acieve this, all children of the stack should be hidden at start.

Fixed. https://github.com/nlogozzo/NickvisionTubeConverter/pull/82/commits/6d0673fc330de57d91e026824e4646390a424c49

nlogozzo:
> I don't like how numbering is implemented

Fixed https://github.com/nlogozzo/NickvisionTubeConverter/pull/82/commits/373169afb7e71626eed9977cd542d319d200338a

> but not to entries in popovers

This I couldn't do however since the numbering is handled by the controller and not via the model, and since editing the entry in the popover changes the model's title it would override the number from the controller. So the number addition is reflected in the entry of the popover, however, I think it still works better and is acceptable this way.

PR #91, opened 2023-03-14
Release fixes

fsobolev:
1. Fixed the problem when the app was starting with 2 windows open and dependency error notification. The cause of problem was that we now have dbus service that starts the app with `--gapplication-service` flag, but the app wasn't able to pass it to `Gio.Application` thus opening second window that, being started by dbus and as result isolated somehow, wasn't able to init python properly. I'm going to bring this changes to Application together with dbus service so we don't worry about it for any app anymore.
2. Set temp dir to be inside config dir on Linux
3. Properly show download progress with unknown total size by using estimated size (I thought I fixed it, but I guess I did changes some night locally and forgot that I didn't commit that 😅 )

### DIFF
--- a/NickvisionTubeConverter.GNOME/Program.cs
+++ b/NickvisionTubeConverter.GNOME/Program.cs
@@ -29,7 +29,7 @@ public partial class Program
     /// </summary>
     /// <param name="args">string[]</param>
     /// <returns>Return code from Adw.Application.Run()</returns>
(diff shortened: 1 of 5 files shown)


nlogozzo:
> I'm going to bring this changes to Application together with dbus service so we don't worry about it for any app anymore.

Thank you :)

PR #110, opened 2023-03-20
Running in the background

### LINKED ISSUE #99: Running in the background
### Discussed in https://github.com/nlogozzo/NickvisionTubeConverter/discussions/96

<div type='discussions-op-text'>

<sup>Originally posted by **fsobolev** March 15, 2023</sup>
With the introduction of queue and playlist support, TC is now very comfortable to use for downloading long playlists. And one more feature that can improve user experience in such cases can be an ability to run the app in the background.
Due to major differences in Windows and GNOME as platforms, this should be implemented very differently for each version (but with some common controllers logic ofc).

## Common

A new setting should be added. When it's disabled, the app will ask for confirmation when a user closes the window while there are downloads running, just like it does now. When enabled, the window will get hidden when user closes it and the rest that will happen depends on platform.
I'm not sure if this should be enabled by default for WinUI, probably yes if it will use system tray, I can't recall any app that has system tray functionality disabled by default. But for GNOME this setting should be disabled by default (explanation below).
The app will show stats when running in the background:
> Downloading: N — P% (S)

Where N is the sum of active and queued downloads, P is total progress in percent, S is total speed.
Because we don't use gettext, we can't use complex plural forms required for a lot of languages, and so something more natural like "Running N download"/"Running N downloads" can't be naturally translated to all languages. Another reason to use proposed string is because it better be as short as possible: in GNOME we will not be able to use string containing more than 96 characters.

## WinUI

I don't know if there are any new patterns in Windows that can be used, so probably we should use the <del>good</del> old system tray icon. Its submenu can contain "Open", "Add Download" and "Quit", and it can show stats in the tooltip. But I'm not aware of all the possibilities in Windows tray, so maybe stats can be shown some better way.
Setting title: "Use system tray"
Description: "Show system tray icon and hide the app when closing the window if there are downloads running."
I think if there are no downloads running, the app should quit on window closing even if the setting is enabled. When downloads are finished while running in the background, the app should stay in tray.

## GNOME

GNOME 44 has [background apps UI](https://www.omglinux.com/gnome-shell-background-apps-ui/). It shows flatpak apps running without a visible window and such apps can set up to 96 characters message to display.
Setting title: "Allow running in the backgound"
Description: "Hide the window instead of quitting if there are downloads running. Showing background activity status requires GNOME 44+"
Background apps UI requires the app to be a flatpak, so the setting should be disabled by default (and GNOME 44+ requirement is one more reason for this). The setting should be hidden and the value in config should be ignored in it's not a flatpak installation, so only the default behavior will be available for non-flatpak. AFAIK we can't check GNOME Shell version in flatpak, so users will be able to make the app hide without any indication of background activity on older GNOME versions, but it's up to a user to decide if such behavior is acceptable for them, and the setting's description warns that GNOME 44+ is required to show background activity.
If while running in the background all downloads are finished and all of them are successful, the app should send notification and quit, background apps UI is not a tray and the app should not just sit in the background doing nothing and waiting for a user to quit it (and if a user enabled the setting when using the old version of GNOME then there would be no indication that it's still running pointlessly). But if some downloads are finished with an error, then the app should continue running, the message text should be changed to something like "Some downloads has failed" and also a notification should contain information about errors, so a user will be aware of them and will be able to open the app, look at logs and retry downloading if they want so.</div>

fsobolev:
Closes #99 

Details:
* New setting added to run in the background. By default, true on Windows and false on Linux (as described in the issue)
* Speed is now saved in bytes per second (like it is provided by yt-dlp) and then both in the window and in background report the speed is converted to more convenient format (see `Shared/Helpers/SpeedFormatter.cs`)
* In GNOME it's disallowed to open multiple windows. The app will quit if all downloads successfully finished in the background.
* Instead of using `hide-on-close` I found it easier and more elegant to manually control window visibility in on close callback
* `if (stage == DownloadStage.Completed && (!GetFocus()!.GetHasFocus() || !GetVisible()))` - to my surprise, for hidden window `GetFocus().GetHasFocus()` returns true, so I also check if window is visible
* Changed notification text in case of an error

TODO:
- [x] WinUI
- [x] Fix total progress calculation (during postprocessing it shows very big numbers, >1000%, scales with number of incomplete downloads)
- [x] (GNOME) Disable the setting when not running in flatpak

Small problem that I don't know if we'll be able to fix: since download progress is shown per file, total progress goes backward when a new file starts downloading. I will later look if we could get from yt-dlp how many files it's going to download.


NickvisionTubeConverter.GNOME/Views/MainWindow.cs
```diff
@@ -32,6 +58,32 @@ private MainWindow(Gtk.Builder builder, MainWindowController controller, Adw.App
...
+                    IntPtr.Zero, // Reply type
+                    0, // Flags
+                    -1, // Timeout
+                    IntPtr.Zero, // Cancellable
+                    IntPtr.Zero, // Callback
+                    IntPtr.Zero); // User data
+                g_variant_builder_unref(builder);
+                return true;
+            }
+            return false;
```
DaPigGuy, line 84:
```suggestion
            }
            return _isBackgroundStatusReported;
```

DaPigGuy (approved):
Looks good so far! :)

NickvisionTubeConverter.Shared/Helpers/SpeedFormatter.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace NickvisionTubeConverter.Shared.Helpers;
+
+public static class SpeedFormatter
+{
+    private static Localizer _localizer;
+
+    static SpeedFormatter()
+    {
+        _localizer = new Localizer();
```
nlogozzo, line 11: `Localizer` is an `IDisposable` and it's hard to dispose this object because it is kept as a static variable and has no object lifetime.

I would not keep the localizer as a static variable here, but instead, make the `GetString` method require the localizer object:
```csharp
public static string GetString(double speed, Localizer localizer) { ... }
```

NickvisionTubeConverter.Shared/Resources/Strings.resx
```diff
@@ -442,4 +478,12 @@
 	<data name="SaveWarning.GTK" xml:space="preserve">
 		<value>The selected save folder is outside the Flatpak sandbox. The file will be saved, but the folder path will not be remembered for future use.&#xA;&#xA;If you want to always remember this folder, give Tube Converter Flatpak write permissions to the selected folder. (You can use Flatseal to achieve this task)</value>
 	</data>
+
+	<data name="AllowBackground.GTK" xml:space="preserve">
+		<value>Allow running in the background</value>
```
nlogozzo, line 483: Could we make these non-GTK only strings? I'd like to implement this switch on WinUI too as most applications allow users to turn off the system-tray icon if they choose.

nlogozzo (changes requested):

nlogozzo:
Ok WinUI implementation done!

> (GNOME) Disable the setting when not running in flatpak

Is this done, or no?

fsobolev:
> Is this done, or no?

Done, the row in preferences is hidden and the config value is ignored if not running in flatpak.

nlogozzo (approved):

Issue #105, opened 2023-03-18
Download only specified timeframe


ElFrogster:
Feature suggestion:
Allowing to specify a timeframe of the video to download might be a useful feature. It could for example be useful for people who only want to download a specific segment of a several hours long video without having to use the bandwidth to download the whole thing.

fsobolev:
https://www.reddit.com/r/youtubedl/wiki/howdoidownloadpartsofavideo/
An example from this article works with CLI, so we'll be able to implement it.

nlogozzo:
What should the UI be? Should it ne another row in the dialog where the user can simply set a start timestamp and an off timestamp and we'll convert it to seconds and pass it to ytdlp?

fsobolev:
> What should the UI be? Should it ne another row in the dialog where the user can simply set a start timestamp and an off timestamp and we'll convert it to seconds and pass it to ytdlp?

Having an additional row for each item in a playlist seems inconvenient to me. I think we can add a button in rows (maybe with `edit-cut` icon) that will open a popover with entries to set timestamps and a label to show errors in case the data is not valid:
1. text doesn't look like timestamp or a number of seconds
2. start timestamp > end timestamp
3. end timestamp > length of media

Also we should disable entries and show a message if a file in the selected format is only available through DASH protocol (can't be used to download part of video).
Take note that this feature works by using ffmpeg as фт external downloader, so #29 should disable partial downloading.

nlogozzo:
https://github.com/NickvisionApps/TubeConverter/releases/tag/2023.6.1-rc1

PR #111, opened 2023-03-20
Choosing specific video resolution

### LINKED ISSUE #100: [Feature Request] Ability to choose specific resolution
Hi!

It would be great to be able to choose a specific resolution. Or at least have an indication, like "Best (4K)", "Good (1080p)" etc.

I often want to download the video in a specific resolution but I never know which resolution the words "Best", "Good", etc. refer to on any given video. Sometimes the highest resolution available for a video is 1080p, sometimes it is 4K but, for most of my use I almost never want to download the 4K version — or any higher than 1080p, actually.

DaPigGuy:
Closes #100 

**TODO**
- [x] Parse available resolutions
- [ ] Pass a user-choosen resolution to Download
- [ ] GNOME UI
- [ ] Windows UI

### DIFF
--- a/NickvisionTubeConverter.Shared/Models/VideoUrlInfo.cs
+++ b/NickvisionTubeConverter.Shared/Models/VideoUrlInfo.cs
@@ -76,6 +76,23 @@ public class VideoInfo : INotifyPropertyChanged
     }
 
     private void NotifyPropertyChanged([CallerMemberName] string propertyName = "") => PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
+
+    /// <summary>
+    /// Parses video info from the Python dictionary returned by yt-dlp
+    /// </summary>
+    /// <param name="videoInfo">The dictionary</param>
+    /// <param name="isPartOfPlaylist">Whether or not the video is part of a playlist</param>
+    /// <returns>A VideoUrlInfo object</returns>
+    internal static VideoInfo ParseFromPyDict(Python.Runtime.PyDict videoInfo, bool isPartOfPlaylist = false)
+    {
+        var title = videoInfo.HasKey("title") ? videoInfo["title"].As<string?>() ?? "Media" : "Media";
+        var url = videoInfo["webpage_url"].As<string>();       
+        foreach (var c in Path.GetInvalidFileNameChars())
+        {
+            title = title.Replace(c, '_');
+        } 
+        return new VideoInfo(url, title, isPartOfPlaylist);
+    }
 }
 
 /// <summary>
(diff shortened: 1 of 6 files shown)


fsobolev:
Would be nice if you could also save info whether the selected format uses DASH protocol, to prepare for #105

NickvisionTubeConverter.Shared/Models/VideoUrlInfo.cs
```diff
@@ -153,4 +152,17 @@ private VideoUrlInfo(string url)
...
         });
     }
+
+    /// <summary>
+    /// Parses video info from the Python dictionary returned by yt-dlp
+    /// </summary>
+    /// <param name="videoInfo">The dictionary</param>
+    /// <param name="isPartOfPlaylist">Whether or not the video is part of a playlist</param>
+    /// <returns>A VideoUrlInfo object</returns>
+    private static VideoInfo ParseVideoInfo(Python.Runtime.PyDict videoInfo, bool isPartOfPlaylist = false)
```
nlogozzo, line 162: Could you make this method a static method of VideoInfo? Something like:

internal static VideoInfo ParseFromPyDict(...)

Then you can call it using VideoInfo.ParseFromPyDict() .

DaPigGuy:
> Would be nice if you could also save info whether the selected format uses DASH protocol, to prepare for https://github.com/nlogozzo/NickvisionTubeConverter/issues/105

It doesn't look like that info is needed anymore, the very first paragraph of the Reddit wiki states that the article is outdated and that yt-dlp has introduced the feature natively

fsobolev:
> It doesn't look like that info is needed anymore, the very first paragraph of the Reddit wiki article states that it is outdated and that yt-dlp has introduced the feature natively. This should also alleviate the need to parse ffmpeg for progress :))

Usual people: read only 1st paragraph
@fsobolev: reads everything but 1st paragraph 🤣 

I just tried it in CLI and yes, it works with DASH, but no, it doesn't use native downloader, utilizing ffmpeg and ignoring `--downloader` flag.

DaPigGuy:
> but no, it doesn't use native downloader, utilizing ffmpeg and ignoring --downloader flag.

Darn... but at least the DASH info doesn't need to be parsed here :)

DaPigGuy:
I will change the selectable file types to audio formats only if no video resolutions are found, although not sure how this should work for playlists, I'm really not sure how many services would have a mix of video and audio-only media...

fsobolev:
> I'm really not sure how many services would have a mix of video and audio-only media...

I doubt there are any. But in this case we can force-uncheck audio-only items in a playlist if the selected target format is video format.

NickvisionTubeConverter.Shared/Models/VideoUrlInfo.cs
```diff
@@ -154,3 +184,32 @@ private VideoUrlInfo(string url)
         });
     }
 }
+
+/// <summary>
+/// A model of an available video resolution
+/// </summary>
+public class VideoResolution
```
nlogozzo, line 191: Sorry to be nit-picking but because `VideoResolution` class is used by `VideoInfo` class, it should be placed on top of that class and thus `VideoResoltion` should be the first class in the file.

DaPigGuy:
How should we handle playlists where videos have varying available resolutions? Should we just show all available resolutions of videos selected for download and automatically choose the closest resolution?

nlogozzo:
@DaPigGuy Could you resolve conflicts and sync with main?

PR #101, opened 2023-03-15
GNOME - Snap fix

fsobolev:
This PR adds workaround to let TC find python `.so` in snap. For this, `TC_PYTHON_SO` is used.

Some notes:
* By default `just` runs command with `set -u` which makes sh exit with an error when unset variable is used. But the output os not very helpful, so I implemented kinda "try-catch" by setting `set +u` and then manually checking if the var is empty.
* It's not for snap only really, if some other package manager will require to use some weird path to python `.so` this var can be used because the app looks for it first and only then uses python to find path if needed.

@soumyaDghosh Please try to build snap with the code from this branch to check that it works and then it we'll be merged.
In YAML file:
```
override-prime: |
      craftctl default
      cd /root/parts/tube-converter/build/NickvisionTubeConverter.GNOME
      TC_PYTHON_SO=/path/to/python/file.so just install-snap
```
Also, I already mentioned it but I will repeat: `dotnet-sdk` is only needed for building and `dotnet-runtime` is not needed too if you use `just publish-self-contained`.

PR #117, opened 2023-03-24
Snap 2023.3.0

fsobolev:
2023.3.0 release with changes from #101 (except for justfile recipe, because it's not needed).


nlogozzo (approved):

fsobolev:
Restored the branch, it will be used in snap

PR #118, opened 2023-03-24
GNOME 44

### LINKED ISSUE #102: GNOME - Use Secondary Windows instead of MessageDialog for dialogs
*Example from Fractal*

DaPigGuy:
https://github.com/nlogozzo/NickvisionApplication/issues/49

Closes #102 

## TODO 
- [x] Basic Implementation
- [x] Move back button to headerbar like in AccountSettingsDialog in Denaro
- [x] Make ScrolledWindow to not use more space than needed

### DIFF
--- a/NickvisionTubeConverter.GNOME/Blueprints/add_download_dialog.blp
+++ b/NickvisionTubeConverter.GNOME/Blueprints/add_download_dialog.blp
@@ -68,11 +68,11 @@ Adw.Window _root {
           Adw.ViewStackPage {
             name: "pageDownload";
             child: Gtk.Box _downloadPage {
-              height-request: 390;
               visible: false;
             
               Gtk.Box {
                 orientation: vertical;
+                hexpand: true;
                 spacing: 6;
 
                 Adw.PreferencesGroup {
@@ -158,7 +158,8 @@ Adw.Window _root {
                 }
 
                 Gtk.ScrolledWindow {
-                  height-request: 200;
+                  propagate-natural-height: true;
+                  max-content-height: 150;
 
                   Adw.PreferencesGroup _videosGroup {
                     margin-start: 2;
--- a/NickvisionTubeConverter.GNOME/Blueprints/add_download_dialog.blp
+++ b/NickvisionTubeConverter.GNOME/Blueprints/add_download_dialog.blp
@@ -54,9 +54,8 @@ Adw.Window _root {
 
           Adw.ViewStackPage {
             name: "pageDownload";
-            child: Gtk.ScrolledWindow _scrollDownload {
+            child: Gtk.Box _downloadPage {
               height-request: 390;
-              hscrollbar-policy: never;
               visible: false;
             
               Gtk.Box {
@@ -160,9 +159,13 @@ Adw.Window _root {
                   }
                 }
 
-                Adw.PreferencesGroup _videosGroup {
-                  margin-start: 2;
-                  margin-end: 2;
+                Gtk.ScrolledWindow {
+                  height-request: 200;
+
+                  Adw.PreferencesGroup _videosGroup {
+                    margin-start: 2;
+                    margin-end: 2;
+                  }
                 }
               }
             };
--- a/NickvisionTubeConverter.GNOME/Blueprints/add_download_dialog.blp
+++ b/NickvisionTubeConverter.GNOME/Blueprints/add_download_dialog.blp
@@ -11,6 +11,19 @@ Adw.Window _root {
 
     Adw.HeaderBar {
       title-widget: Gtk.Label {};
+
+      [start]
+      Gtk.Button _backButton {
+        halign: start;
+        valign: center;
+        visible: bind _downloadPage.visible;
+
+        Adw.ButtonContent {
+          icon-name: "go-previous-symbolic";
+          label: _("Back");
+        }
+      }
+
       styles ["flat"]
     }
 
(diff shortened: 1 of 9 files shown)


DaPigGuy:
@fsobolev Could you update the Add Download dialog screenshot

NickvisionTubeConverter.GNOME/Blueprints/add_download_dialog.blp
```diff
@@ -1,150 +1,187 @@
...
+                    }
+                  }
 
-            Adw.ButtonContent {
-                icon-name: "view-list-ordered-symbolic";
-                label: _("NumberVideos");
+                  Adw.PreferencesGroup _videosGroup {
+                    margin-start: 2;
+                    margin-end: 2;
+                  }
```
fsobolev, line 168: This should be wrapped in Gtk.ScrolledWindow, otherwise for long playlists the window will be incredibly big.

fsobolev (changes requested):

NickvisionTubeConverter.GNOME/Blueprints/add_download_dialog.blp
DaPigGuy: The Adw.ViewStack is already scrollable oddly enough...

fsobolev: > ~The Adw.ViewStack contents are already scrollable oddly enough...~
> 
> I've just noticed that the child widget for the download page is a Gtk.ScrolledWindow

Right, I missed that. But we don't want the whole page to be scrollable. At least the heading and the download button should always be visible to a user.

nlogozzo: > I see the button is outside of Gtk.ScrolledWindow. Still, I think it would be best to make only videos list scrollable.

Agreed. The videos list should be the only thing scrollable.

DaPigGuy: Done, feel free to tweak

nlogozzo:
@fsobolev Can we remove `AddDownloadDialogController.Accepted`?

DaPigGuy:
Yes, it's not used anywhere

nlogozzo (approved):

nlogozzo:
LGTM!

fsobolev:
| Playlist | Single video |
| --- | --- |
| Снимок экрана от 2023-03-25 05-30-01 | Снимок экрана от 2023-03-25 05-28-13 |

In both cases, doesn't fit my screen (1366x768).
Suggestions:
* Make ScrolledWindow to not use more space than needed
* Remove heading with media/playlist title and move back button to headerbar like I did in AccountSettingsDialog in Denaro
* Think about styling of the list. The first thing that I got in mind is to add margins to PreferencesGroup and add "card" class to ScrolledWindow, but maybe we can make it look better.

Снимок экрана от 2023-03-25 05-32-14

nlogozzo:
> move back button to headerbar like I did in AccountSettingsDialog in Denaro

Agreed. 

> Remove heading with media/playlist title

I don't know about this...I feel like it good to have the title of the playlist so the user remembers where they are downloading from and for individual videos if a user changes the filename they can still see the original title.

> Make ScrolledWindow to not use more space than needed

Agreed. 

> Think about styling of the list. The first thing that I got in mind is to add margins to PreferencesGroup and add "card" class to ScrolledWindow, but maybe we can make it look better.

I don't know how I feel about adding `card` to the scrolledwindow...i think it looks great the way it is without it. Of course if we find a better design all together, I'm open to that.

DaPigGuy:
> I don't know how I feel about adding card to the scrolledwindow...i think it looks great the way it is without it. Of course if we find a better design all together, I'm open to that.

Agreed

DaPigGuy:
There seems to be a problem with b1f5e1ec7911ad8785c1b206907fb3d4c74ec87d where the scroll window does not expand when a video row's title gets long enough, despite being far smaller than 150 in height

fsobolev:
> I don't know how I feel about adding `card` to the scrolledwindow...i think it looks great the way it is without it. Of course if we find a better design all together, I'm open to that.

The problem is that the playlist gets "chopped off" without a visible borders of the area it is limited to, kinda a piece of the list is floating above the window... I really don't like it. But `card` has some other downsides too.
So instead I implemented this:

| Scrollable | Not scrollable |
| --- | --- |
| Снимок экрана от 2023-03-25 16-15-17 | Снимок экрана от 2023-03-25 16-14-34 |

Borders and margins are only added when the list is scrollable. This requires sizing to properly work, so I set `title-lines` to 1 as I suggested above.

nlogozzo:
Looks 100x better!

nlogozzo:
Looks 100x better!

nlogozzo:
Anything left to do?

fsobolev:
> Anything left to do?

Everything's good to me. If @DaPigGuy doesn't want to change something more, then it's good to merge.

fsobolev (approved):

nlogozzo (approved):

nlogozzo:
@DaPigGuy let me know when to merge..or merge yourself if it lets you :)

DaPigGuy:
I find the borders a little odd, but have no better ideas, otherwise LGTM

nlogozzo:
https://github.com/nlogozzo/NickvisionTubeConverter/releases/tag/2023.3.1

Issue #134, opened 2023-03-29
change "url" strings value to uppercase "URL"


Krindog7337:
because URL is an abbreviation from Uniform Resource Locator, then it should be URL/URLs not Url
![image](https://user-images.githubusercontent.com/92032579/228559985-bcf53f87-3cfe-497c-a911-26a1d4d8128b.png)

nlogozzo:
https://github.com/nlogozzo/NickvisionTubeConverter/releases/tag/2023.4.0

PR #135, opened 2023-03-29
change url strings value to URL

Krindog7337:
Issues #134

### DIFF
--- a/NickvisionTubeConverter.Shared/Resources/Strings.resx
+++ b/NickvisionTubeConverter.Shared/Resources/Strings.resx
@@ -366,15 +366,15 @@
 
 	<!--Fields-->
 	<data name="VideoUrl.Field" xml:space="preserve">
-		<value>Video Url</value>
+		<value>Media URL</value>
 	</data>
 
 	<data name="VideoUrl.Invalid" xml:space="preserve">
-		<value>Video Url (Invalid)</value>
+		<value>Media URL (Invalid)</value>
 	</data>
 	
 	<data name="VideoUrl.Placeholder" xml:space="preserve">
-		<value>Enter video url here</value>
+		<value>Enter media URL here</value>
 	</data>
 
 	<data name="FileType.Field" xml:space="preserve">


nlogozzo (approved):

fsobolev:
I think it should be changed from "Video URL" to "Media URL" or "Video, audio or playlist URL". As I said in some other PR and in matrix, it would be better to give users more hints that the app is capable of downloading not only youtube videos, but everything yt-dlp can get. https://github.com/nlogozzo/NickvisionTubeConverter/discussions/121#discussion-5004266 is a good evidence that "Video URL" can be confusing.

Krindog7337:
> I think it should be changed from "Video URL" to "Media URL" or "Video, audio or playlist URL". 

> it would be better to give users more hints that the app is capable of downloading not only youtube videos, but everything yt-dlp can get.

in that case, how about giving a pop-over info on right side of text input? like on pointer assistance settings (GNOME 44)

image
image

NickvisionTubeConverter.Shared/Resources/Strings.de.resx
```diff
@@ -334,7 +334,7 @@ marcin https://github.com/martin-desktops</value>
     <value>Wenn diese Option aktiviert ist, werden die Video-Metadaten in die heruntergeladene Datei aufgenommen.</value>
   </data>
   <data name="VideoUrl.Invalid" xml:space="preserve">
-    <value>Video Url (ungültig)</value>
+    <value>Video URL (ungültig)</value>
```
fsobolev, line 337: You don't need to change anything in Strings.*.resx, Weblate will look for all changes in original Strings.resx and make all translation files match.

fsobolev (changes requested):

fsobolev:
> in that case, how about giving a pop-over info on right side of text input? like on pointer assistance settings (GNOME 44)

I like it, will implement it later 👍🏻

fsobolev (approved):

PR #140, opened 2023-03-31
GNOME - Add Validate URL button

### LINKED ISSUE #139: The button to verify URLs is not intuitive
## Problem

Attempting to download a YouTube video

When I paste a YouTube link to "Video URL" input box, I was expecting to download button to become clickable but it didn't which made me confused.

I see the verify button, but at first,  I thought it was a badge, not a button.

### Ideal design

Automatically verify the URL on paste.

### Alternative design

The verify button should be as big and visible as the download button. There is no reason to show a download button when URL hasn't been verified. Once the user clicks the big verify button (and it passes), the verify button should be replaced by the download button.

fsobolev:
[tc-validate.webm](https://user-images.githubusercontent.com/117388856/229248604-247d6532-546e-4cbe-9469-421736114ed8.webm)

Not sure whether "Validate URL" is the best label, but I don't have a better idea.
Also fixed playlist preferences group margins (can be noticed on the video that it was cut off on top and bottom).
Closes #139

### DIFF
--- a/NickvisionTubeConverter.GNOME/Blueprints/add_download_dialog.blp
+++ b/NickvisionTubeConverter.GNOME/Blueprints/add_download_dialog.blp
@@ -51,7 +51,6 @@ Adw.Window _root {
                 Adw.EntryRow _urlRow {
                   title: _("VideoUrl.Field");
                   width-request: 420;
-                  show-apply-button: true;
                 }
               }
 
@@ -60,6 +59,16 @@ Adw.Window _root {
                 width-request: 40;
                 height-request: 40;
                 visible: false;
+              }
+
+              Gtk.Button _validateUrlButton {
+                label: _("ValidateUrl");
+                halign: center;
+                valign: center;
+                margin-top: 24;
+                margin-bottom: 24;
+
+                styles ["pill", "suggested-action"]
               }
             };
           }
@@ -163,18 +172,18 @@ Adw.Window _root {
                   margin-end: 24;
                 }
               }
+
+              Gtk.Button _addDownloadButton {
+                label: _("Download");
+                halign: center;
+                valign: center;
+                margin-top: 24;
+                margin-bottom: 24;
+
+                styles ["pill", "suggested-action"]
+              }
             };
           }
-        }
-
-        Gtk.Button _addDownloadButton {
-          label: _("Download");
-          halign: center;
-          valign: center;
-          margin-top: 24;
-          margin-bottom: 24;
-
-          styles ["pill", "suggested-action"]
         }
       }
     }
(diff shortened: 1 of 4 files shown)


nlogozzo:
Maybe just say "Validate"

nlogozzo (approved):

KSXGitHub:
This is much better.

But I think auto verification can still work in some limited fashion. Such as: It will only auto verify when user **paste** an URL, it will not auto verify if user merely type the URL.

fsobolev:
> Such as: It will only auto verify when user paste an URL, it will not auto verify if user merely type the URL.

The widget doesn't have a signal to determine when the text was pasted from clipboard.

KSXGitHub:
Oh, I see.

Some torrent application such as Transmission can detect magnet link in the clipboard. When I copy a magnet link and then click "Open magnet link" in Transmission, the URL box already has the magnet link prefilled.

Maybe Tube Converter can work in similar way? If the clipboard already has a YouTube URL, clicking "Add Download" should have the URL prefilled and auto validate.

fsobolev:
> Some torrent application such as Transmission can detect magnet link in the clipboard. When I copy a magnet link and then click "Open magnet link" in Transmission, the URL box already has the magnet link prefilled.
> 
> Maybe Tube Converter can work in similar way? If the clipboard already has a YouTube URL, clicking "Add Download" should have the URL prefilled and auto validate.

Generally I like this idea.

But I don't think it should be auto-validated, this can be irritating. Imagine a situation if a user copies a url, opens Tube Converter and when the link appears in the dialog they realize that this link points to a video and playlist (`youtube.com/watch?v=video_id&list=playlist_id`). This link will start a process of retrieving the data for the whole playlist which can be long. So the user will need to close the dialog, get proper link without playlist id and open the dialog again.
And, well, I think it just feels wrong when an app you use provides you some visual controls that are activated automatically.

Also the question is should the content of clipboard be pasted if there's a youtube link or any link, because yt-dlp (that Tube Converter uses under the hood) supports *hundreds* of sites.

nlogozzo:
> And, well, I think it just feels wrong when an app you use provides you some visual controls that are activated automatically without your interaction.

Agreed.

> Also the question is should the content of clipboard be pasted if there's a youtube link or any link, because yt-dlp (that Tube Converter uses under the hood) supports hundreds of sites.

I believe we should make it where if the most recent entry in the clipboard is a URL, we can just automatically paste it in the entry and then the user will click validate. Maybe we should have a settings in Preferences to turn the auto-pasting off...

fsobolev:
> I believe we should make it where if the most recent entry in the clipboard is a URL, we can just automatically paste it in the entry and then the user will click validate. 

So any url (`http://...`, `https://...`)? Okay

> Maybe we should have a settings in Preferences to turn the auto-pasting off...

Agreed

nlogozzo:
> So any url (http://..., https://...)? Okay

Yes, as long as we have the option to turn it off in preferences it'll be ok.

PR #148, opened 2023-04-03
Update org.nickvision.tubeconverter.json

nlogozzo:


nlogozzo:
Hmmm failed here too...https://github.com/flathub/org.nickvision.tubeconverter/pull/41
oh well

PR #149, opened 2023-04-03
Unescape filename and remove save warning (flatpak)

### LINKED ISSUE #141: Unnecessary blackslash in resulting filename
When I use Tube Converter to download [WE ARE JAPANESE GOBLIN [MMD]](https://www.youtube.com/watch?v=UIp6_0kct_U), the resulting file is named `WE ARE JAPANESE GOBLIN \[MMD\].mp4` instead of `WE ARE JAPANESE GOBLIN [MMD].mp4`. I can rename the file of course, but it would be super annoying if I was to download multiple videos and then have to rename all of them afterward.

fsobolev:
Closes #141


nlogozzo:
Can you remove the string `SaveWarning.GTK` from the resources file since it is no longer used?

nlogozzo (approved):

PR #181, opened 2023-04-12
Logging, unescaping and overwriting improvements

### LINKED ISSUE #152: Unescape filenames in notifications and in logs
Since fixing #141 filenames are now unescaped after downloading, but they are still shown escaped in notifications, and there's no message in logs that unescaping happened.

### LINKED ISSUE #161: Logs are not completely shown in the app
The first character on the first line is not shown (`[`), and last lines in case of an error are not shown either.

Example. In the file:
```
[ZDF] Extracting URL: https://www.zdf.de/show/mai-think-x-die-show/maithink-x-folge-24-100.html
...
[download] 100% of    1.12GiB in 00:03:04 at 6.24MiB/s   
[VideoRemuxer] Remuxing video from mp4 to webm; Destination: /var/home/user/.var/app/org.nickvision.tubeconverter/config/Nickvision/Nickvision Tube Converter/temp/518ab231-5087-4789-92ad-2c9f3c533bc3/Der\ gefährliche\ Naturtrend\.webm
ERROR: Postprocessing:   Stream #0:1 -> #0:1 (copy)
```

In the app:
```
ZDF] Extracting URL: https://www.zdf.de/show/mai-think-x-die-show/maithink-x-folge-24-100.html
...
[download] 100% of    1.12GiB in 00:03:04 at 6.24MiB/s
```

fsobolev:
* Don't clean temp dir to avoid redownloading files from the beginning in case the download was restarted
* Check if final file already exists when `_overwriteFiles` is false to prevent download that can't in fact be successfully finished
* Unescape filename on error, closes #152
* Improve logging, closes #161 
    There were 3 problems with logs:
    1. The first character was not shown
    2. Logs weren't completely shown in the app in case of error (because we only updated logs from the hook, now it's also updated on error)
    3. Logs are not always in sync with what is actually happening. This is not fixed. It's odd, yt-dlp calls hook for download after it prints the output, but for postprocessing it's the opposite. In result, `progressCallback` is called first, and only then the file with the log is changed, so it will be shown on next hook call. I was thinking how to workaround this, but tbh I think it doesn't worth the effort, the purpose of logs is to see what happened on error, and in this case the full log is shown as I said earlier
* EmbedThumbnail postprocessor replaced with custom one that ignores errors (similar to TCEmbedMetadata)

### DIFF
--- a/NickvisionTubeConverter.GNOME/Blueprints/download_row.blp
+++ b/NickvisionTubeConverter.GNOME/Blueprints/download_row.blp
@@ -169,7 +169,9 @@ Adw.Bin _root {
           margin-end: 6;
           margin-bottom: 6;
           halign: start;
+          valign: start;
           wrap: true;
+          wrap-mode: char;
           selectable: true;
         };
 
--- a/NickvisionTubeConverter.GNOME/Controls/DownloadRow.cs
+++ b/NickvisionTubeConverter.GNOME/Controls/DownloadRow.cs
@@ -36,15 +36,15 @@ public partial class DownloadRow : Adw.Bin, IDownloadRowControl
 
     private readonly Localizer _localizer;
     private readonly Download _download;
+    private readonly GSourceFunc _runStartCallback;
+    private readonly GSourceFunc _runEndCallback;
+    private readonly GSourceFunc _stopCallback;
+    private readonly GSourceFunc _updateLogCallback;
+    private readonly GSourceFunc _processingCallback;
+    private GSourceFunc? _downloadingCallback;
     private bool? _previousEmbedMetadata;
     private bool _wasStopped;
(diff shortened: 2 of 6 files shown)


nlogozzo:
Don't worry about `Codacy Static Code Analysis`. If the PR is ready, I'll merge.

fsobolev:
> Don't worry about `Codacy Static Code Analysis`. If the PR is ready, I'll merge.

I'm trying to figure out what I can do with EmbedThumbnail postprocessor, it rarely but randomly fails, and even keeping temp folder didn't help to prevent restarting the download from the beginning, because postprocessors that are called before EmbedThumbnail remove originally downloaded files... Probably I will do something similar to Metadata postprocessor. Then the PR will be ready :)

nlogozzo (approved):

PR #183, opened 2023-04-12
Speed Limit

### LINKED ISSUE #174: Option for maximum download speed
Currently when downloading videos, Tube Converter appears to be using most of my bandwidth, making streaming videos at the same time nearly unusable. An option to limit the download speed would be useful, especially when running the program in the background.

fsobolev:
Closes #174 

| Add Download Dialog | Preferences Window |
| --- | --- |
| Снимок экрана от 2023-04-12 15-33-54 | Снимок экрана от 2023-04-12 15-34-48 |

Speed limit should be in range 512-10240 KiB/s, I think these are sane values. The limit is per-download, so for example 3 downloads with 512KiB/s limit will use ~1.5MiB/s in total.


nlogozzo:
WinUI:
image
image

nlogozzo (approved):

PR #203, opened 2023-04-24
GNOME - Progress bar in launcher using libunity

fsobolev:
Unity DE is not as popular as before, but its [launcher API](https://wiki.ubuntu.com/Unity/LauncherAPI) is working in GNOME with some extensions like Dash to Dock and Dash to Panel. With libunity I made Tube Converter show progress bar in launcher.

libunity is an optional dependency, the app will not crash if the lib isn't found.

To build libunity in flatpak I looked at Fedora and Arch packages, I got patches from them. Everything in `NickvisionTubeConverter.GNOME/flatpak` directory should be copied to flathub repo and paths to patches should be changed accordingly in the manifest. `dee` uses simple build system because I didn't find a way to call `autoreconf` another way (without this the build will fail because it will look for old `aclocal-1.13` and fail).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "NickvisionTubeConverter.GNOME/flatpak/shared-modules"]
+	path = NickvisionTubeConverter.GNOME/flatpak/shared-modules
+	url = https://github.com/flathub/shared-modules.git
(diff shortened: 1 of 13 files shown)


nlogozzo (approved):

nlogozzo:

PR #212, opened 2023-04-28
Stop All Downloads / Retry Failed Downloads

### LINKED ISSUE #208: Add StopAllDownloads Action

nlogozzo:
Closes #208

### DIFF
--- a/NickvisionTubeConverter.Shared/Models/Download.cs
+++ b/NickvisionTubeConverter.Shared/Models/Download.cs
@@ -155,7 +155,6 @@ public class Download
         {
             using (Python.Runtime.Py.GIL())
             {
-                _pid = Python.Runtime.PythonEngine.GetPythonThreadID();
                 dynamic ytdlp = Python.Runtime.Py.Import("yt_dlp");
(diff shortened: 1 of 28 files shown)


NickvisionTubeConverter.Shared/Models/Download.cs
```diff
@@ -344,6 +343,7 @@ private void ProgressHook(Python.Runtime.PyDict entries)
         {
             using (Python.Runtime.Py.GIL())
             {
+                _pid = Python.Runtime.PythonEngine.GetPythonThreadID();
```
fsobolev, line 346: Why moved this to progress hook?

fsobolev (changes requested):

NickvisionTubeConverter.Shared/Models/Download.cs
nlogozzo: Once yt-dlp was running i was getting a different thread id when ffmpeg started. Implementing it like this will now also stop ffmpeg conversion whereas before ffmpeg was still running in the background.

fsobolev (approved):

PR #215, opened 2023-04-28
Added the snap files in this repo

soumyaDghosh:
What will be the benefits after adding the snap files?
1. I'll connect this repo with the snapcraft builder, so you guys can simply update the app whenever there is one. There'll be no more delay in snaps updates.
2. People will no longer get confused about creating issues. This will be the same place for all. A `snap` label can also be created for that reason may be.

Disadvantages:
1. The builder may get triggered whenever there is a push in this repo. Which will always create updates for snap in the edge channel. But, the updates will be from the source-tag that we've desired to have and if I don't promote them, they'll not be stable. I am not sure about this though. I have never tested this.
2. I don't know if you guys want to make this snap support official now.

What ever happens I'll be there to maintain the snaps. This will just fast the snap update.


nlogozzo:
I'd be happy to add you as a contributor to this repo and Denaro as well so you have access and can manage the files on the repos as needed.

soumyaDghosh:
I already have the necessary access. Thanks for the proposal though. :+1:

nlogozzo:
I added you as a collaborator with write permissions to the repo itself, that way you can create branches and PR from within this repo without needing to fork and use your own. This will also cause the WinUI CI to not fail.

soumyaDghosh:
Thanks a lot. That's my pleasure.

fsobolev (changes requested):
I think snap folder should be in Nickvision*.GNOME, not in root

soumyaDghosh:
Let me check first if it'll work or not.

soumyaDghosh:
No, as you can see from the commits itself, the snap builder can find files only under folder like snap. Others don't work

fsobolev (approved):

Issue #189, opened 2023-04-13
Crashes with Good and Best quality


fsobolev:
Hi, I still have the same issue using 2023.4.1 (flatpak)

              One addition maybe is quality option.

              It crashes with Good and Best, but has no problem when choosing Worst - then everything works correctly.
              Reproducible: always.

_Originally posted by @ekasprzak in https://github.com/nlogozzo/NickvisionTubeConverter/issues/130#issuecomment-1506858233_

fsobolev:
@ekasprzak Could you please specify when exactly crash happens: immediately after download started, when it finishes or at some other point?
Also console output will really help.
```
flatpak run org.nickvision.tubeconverter
```
And would be nice if you can check whether the problem still exists in [latest beta](https://github.com/nlogozzo/NickvisionTubeConverter/releases/tag/2023.4.2-beta1).

nlogozzo:
I don't know why this would be happening when we wrapped everything in `g_main_context_invoke`...maybe I missed something...

fsobolev:
> I didn't want to open a new issue as the problem is the same as in #130, (I just added some additional info about quality).
> Same in description - "UI hangs and then closes itself" and exactly the same output from console:

Hm... Hard for me to imagine how it can depend on quality, that's why I thought it's a separate issue.

ekasprzak:
> > I didn't want to open a new issue as the problem is the same as in #130, (I just added some additional info about quality).
> > Same in description - "UI hangs and then closes itself" and exactly the same output from console:
> 
> Hm... Hard for me to imagine how it can depend on quality, that's why I thought it's a separate issue.

Yeah, I  get it.
 Just to be fully confirm it, I triple checked it:

One, the same youtube video:
3 attempts with Best quality - all crashed with above error
3 attempts with Good quality - all crashed with above error
3 attempts with Worst quality - all success

nlogozzo:
> One, the same youtube video:

Could you send us the link of the video you were trying to download?

> And would be nice if you can check whether the problem still exists in [latest beta](https://github.com/nlogozzo/NickvisionTubeConverter/releases/tag/2023.4.2-beta1).

And yes, if you could try with the latest beta too see if it still happens.

MetalDroid:
> @ekasprzak Could you please specify when exactly crash happens: immediately after download started, when it finishes or at some other point? Also console output will really help.
> 
> ```
> flatpak run org.nickvision.tubeconverter
> ```
> 
[...]

PR #217, opened 2023-04-30
GNOME - Better move/delete row callbacks

fsobolev:
Related: #189 
Hopefully fix crashes mentioned by mtsk in matrix chat

### DIFF
--- a/NickvisionTubeConverter.GNOME/Helpers/RowCallbackData.cs
+++ b/NickvisionTubeConverter.GNOME/Helpers/RowCallbackData.cs
@@ -1,0 +1,64 @@
+using NickvisionTubeConverter.GNOME.Controls;
+using System;
+using System.Runtime.InteropServices;
+
+namespace NickvisionTubeConverter.GNOME.Helpers;
+
+/// <summary>
+/// An object to pass to move/delete row callbacks in main window
+/// </summary>
+public class RowCallbackData : IDisposable
+{
+    private bool _disposed;
+
+    /// <summary>
+    /// Row to move/delete
+    /// </summary>
+    public DownloadRow Row { get; init; }
+    /// <summary>
+    /// Box to modify
+    /// </summary>
+    public Gtk.Box Box { get; init; }
+    /// <summary>
+    /// GCHandle to pass to unmanaged code
+    /// </summary>
+    public GCHandle Handle { get; init; }
+
+    /// <summary>
+    /// Constructs RowCallbackData
+    /// </summary>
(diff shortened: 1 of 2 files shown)


NickvisionTubeConverter.GNOME/Helpers/RowCallbackData.cs
```diff
@@ -0,0 +1,64 @@
...
+/// An object to pass to move/delete row callbacks in main window
+/// </summary>
+public class RowCallbackData : IDisposable
+{
+    private bool _disposed;
+
+    /// <summary>
+    /// Row to move/delete
+    /// </summary>
+    public DownloadRow Row;
```
nlogozzo, line 17: public variables should be properties similar to `Handle`.

Should be:
```csharp
public DownloadRow Row { get; init; }
public Gtk.Box Box { get; init; }
```

nlogozzo (changes requested):

NickvisionTubeConverter.GNOME/Helpers/RowCallbackData.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// </summary>
+    public DownloadRow Row { get; init; }
+    /// <summary>
+    /// Box to modify
+    /// </summary>
+    public Gtk.Box Box { get; init; }
+    /// <summary>
+    /// GCHandle to pass to unmanaged code
+    /// </summary>
+    public GCHandle? Handle { get; init; }
```
nlogozzo, line 25: You can get rid of the `?` here. The `?` means "this object could be null" but because you have `Handle = GCHandle.Alloc(this)` in the constructor, we know that this object will NEVER be null, thus the `?` is not needed.

This will allow you to remove the `!` from 363 and 374 in `MainWindow.cs`, because the `!` tells the compiler "I know this object can be null but I'm telling you it's not going to be null". But because we remove the `?` meaning the object can never be null, the `!` makes no sense here.

nlogozzo (changes requested):

NickvisionTubeConverter.GNOME/Helpers/RowCallbackData.cs
fsobolev: I kept `(RowCallbackData)(GCHandle.FromIntPtr(dataHandle).Target!)` because otherwise there's a warning about casting nullable type to non-nullable, everywhere else `!` is removed

nlogozzo (approved):

nlogozzo:

PR #220, opened 2023-04-30
GNOME Circle Fixes

### LINKED ISSUE #211: GNOME Circle Fixes
https://gitlab.gnome.org/Teams/Circle/-/issues/128
- AddDownloadDialog
    - > The popover for renaming the downloaded file is weird, this needs a different pattern. Perhaps replacing the row with an entry row would work?
    - > There's too much vertical spacing between the entry and the button on the Add dialog, I'd reduce it to maybe 24 or 30px
        - image
    - > It's unclear what "Speed Limit" is, this needs context. I'd show what the current limit is, and add a reference that this can be changed in preferences in the description. Also, rows with a switch should always have the thing that's being turned on as the label, i.e. "Speed Limit" rather than "Enable Speed Limit". The verb is implicit in the switch if what the row does is just "Turn X on/off".

- DownloadRow
    - >The row layout looks a bit cramped, making it feel off-style. I'd try increasing the outer padding on the row. The small, left-aligned label in the bottom left also doesn't help, I'd maybe get rid of that or move it inline with the URL (or perhaps the URL could be hidden by default, not sure it needs to be always visible on the list rows).
        - Replace URL with status as we did in WinUI (#207)
        - > The large symbolic in the corner looks odd, I'd make it the regular icon size and give it a circle background, similar to e.g. Fragments.

    - > The card-within-a-card styling here doesn't work, I'd drop the card styling on the inner box, and give it a simple border, as well as a smaller radius so it feels harmonious with the larger outer radius.
        - image

- HomePage
    - > The "Good Morning" message on the welcome screen is weird. I could see this maybe working for apps where your daily routine is part of the workflow (e.g. a todo app based around daily tasks), but for a domain-specific utility like this it doesn't make sense IMO. I'd go with a more standard status page with a title like "Download Videos" and the app icon.
    - > More generally, the empty state could be simpler, I'd drop the Add button on the left and the title in the headerbar, and make it flat (i.e. there'd just be a primary menu button in the top right)
        - Done in #212  

- SettingsPage
    - A dropdown is not an appropriate pattern for choosing a numeric value like this, this should probably use the new spin row widget (or just an old-style spinbutton inside a list row, same as the speed limit thing below)
        - image
        - Done in WinUI (#207)

nlogozzo:
Closes #211 

# TODO
## AddDownloadDialog
- [x] The popover for renaming the downloaded file is weird, this needs a different pattern. Perhaps replacing the row with an entry row would work?
- [x] There's too much vertical spacing between the entry and the button on the Add dialog, I'd reduce it to maybe 24 or 30px
- [x] It's unclear what "Speed Limit" is, this needs context. I'd show what the current limit is, and add a reference that this can be changed in preferences in the description. Also, rows with a switch should always have the thing that's being turned on as the label, i.e. "Speed Limit" rather than "Enable Speed Limit". The verb is implicit in the switch if what the row does is just "Turn X on/off".
## DownloadRow
- [x] The row layout looks a bit cramped, making it feel off-style. I'd try increasing the outer padding on the row. The small, left-aligned label in the bottom left also doesn't help, I'd maybe get rid of that or move it inline with the URL (or perhaps the URL could be hidden by default, not sure it needs to be always visible on the list rows).
    - [x] Replace URL with status
    - [x] The large symbolic in the corner looks odd, I'd make it the regular icon size and give it a circle background, similar to e.g. Fragments.
- [x] The card-within-a-card styling here doesn't work, I'd drop the card styling on the inner box, and give it a simple border, as well as a smaller radius so it feels harmonious with the larger outer radius.
## HomePage
- [x] The "Good Morning" message on the welcome screen is weird. I could see this maybe working for apps where your daily routine is part of the workflow (e.g. a todo app based around daily tasks), but for a domain-specific utility like this it doesn't make sense IMO. I'd go with a more standard status page with a title like "Download Videos" and the app icon.
## SettingsPage
- [x] A dropdown is not an appropriate pattern for choosing a numeric value like this, this should probably use the new spin row widget (or just an old-style spinbutton inside a list row, same as the speed limit thing below)


nlogozzo:
See #222

PR #223, opened 2023-05-01
GNOME - Disable filesystem permissions

fsobolev:
This PR removes permissions for user folders and adds handling of inaccessible paths for save folder. `gtk_file_launcher_open_containing_folder` doesn't open real path, but I think it's the best we can use, because the file will be selected in file manager and also, from my experience, portal folders are like mirrors of real folders, moving files from there to another place affects real folder too, so the only issue is confusing path in file manager.


nlogozzo (approved):

nlogozzo:

PR #227, opened 2023-05-02
New Backend

### LINKED ISSUE #222: Stopping/Retrying Multiple Downloads Not Working --> NEW BACKEND
Working with multiple downloads in terms of stopping and retrying them do not work 100% correctly. 
To solve this, I will be writing a new backend altogether. This will not only fix this issue but it will also fix some architectural design choices that I do not like, such as the actual downloads being managed by the UI rows themselves instead of the backend.

nlogozzo:
Closes #222


nlogozzo:
Ok so with the current state of this PR, stopping all downloads and then retrying downloads should be fixed, if you'd like to test it.

If all works, I will probably finish the rewrite here.

nlogozzo:
Ok new backend implemented for both GNOME and WinUI. If you could please test it out and let me know how it goes :)

fsobolev (approved):

PR #239, opened 2023-05-10
GNOME - Add pulsing bar for processing

fsobolev:
@nlogozzo The solution that you suggested with while loop doesn't work well because the bar shouldn't pulse every frame. I think that using a separate bar is the most reliable solution.

### DIFF
--- a/NickvisionTubeConverter.GNOME/Controls/DownloadRow.cs
+++ b/NickvisionTubeConverter.GNOME/Controls/DownloadRow.cs
@@ -15,6 +15,7 @@ namespace NickvisionTubeConverter.GNOME.Controls;
 public partial class DownloadRow : Adw.Bin, IDownloadRowControl
 {
     private delegate void GAsyncReadyCallback(nint source, nint res, nint user_data);
+    private delegate bool GSourceFunc(nint user_data);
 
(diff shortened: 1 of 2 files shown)


NickvisionTubeConverter.GNOME/Controls/DownloadRow.cs
```diff
@@ -98,6 +103,12 @@ private DownloadRow(Gtk.Builder builder, Guid id, string filename, string saveFo
             _lblLog.GetClipboard().SetText(_lblLog.GetText());
             _sendNotificationCallback(new NotificationSentEventArgs(_localizer["LogCopied"], NotificationSeverity.Informational));
         };
+        _pulsingBarCallback = (data) =>
+        {
+            _pulsingBar.Pulse();
+            return true;
```
nlogozzo, line 109: I don't like that this returns true always, causing this callback to run forever every 30 seconds, never stopping....

Can you call g_timeout_add or g_idle_add everytime SetProgressState is called and return false instead?

nlogozzo (changes requested):

NickvisionTubeConverter.GNOME/Controls/DownloadRow.cs
fsobolev: Added a variable to control whether the callback should be running

nlogozzo (approved):

PR #240, opened 2023-05-10
Video resolutions, thumbnail crop, specified timeframe

### LINKED ISSUE #100: [Feature Request] Ability to choose specific resolution

### LINKED ISSUE #105: Download only specified timeframe

### LINKED ISSUE #237: Add option to crop thumbnail to square
Would it be possible to add an option either in the settings or in the download dialog to crop the thumbnail to a square? yt-dlp has that as seen [here](https://github.com/yt-dlp/yt-dlp/issues/429).
It would be especially useful for music coming from YouTube Music

fsobolev:
Closes #100 
Closes #237 
Closes #105 

- [ ] Shared
  - [x] Parse video resolutions
  - [ ] Option to crop thumbnail
  - [ ] Option to specify timeframe
- [ ] Update GNOME
- [ ] Update WinUI


fsobolev:
I reverted changes in top part of AddDownloadDialog because:
- Height is not such a problem now
- (not implemented yet) "Quality" changes to "Maximum Quality" when downloading a playlist, this label will not fit a narrow box
- Circle reviewers will probably not like that unusual layout

| Download playlist | Playlist page | Download single media |
| --- | --- | --- |
| Снимок экрана от 2023-05-11 04-45-14 | Снимок экрана от 2023-05-11 04-45-27 | Снимок экрана от 2023-05-11 04-47-31 |

nlogozzo:
Maybe for the playlist with showing `1 to 2 items` we can add the playlist title somewhere in the dialog to signify a collection of videos are being downloaded.

fsobolev:
> Maybe for the playlist with showing `1 to 2 items` we can add the playlist title somewhere in the dialog to signify a collection of videos are being downloaded.

I think media titles in playlist view are enough for a user to make sure that the correct URL was used. And I can't find a place where the title would look good. "TITLE: X of Y items" doesn't work because if a title is very long then really important part (number of items to download) will get hidden, or the row will have weirdly increased height.
I tried to add a title on playlist page, but it feels odd to me, because the playlist title has nothing to do with the "number titles" button and IMHO the title itself doesn't give the feeling "oh, that's playlist title" from the first glance, it's some element that is unexpected and can lead to confusion, and as I said before I think it's pointless.

image

NickvisionTubeConverter.Shared/Models/MediaUrlInfo.cs
```diff
@@ -0,0 +1,122 @@
...
+            {
+                var resolution = new VideoResolution(format["width"].As<int>(), format["height"].As<int>());
+                if (!VideoResolutions.Exists(r => r == resolution))
+                {
+                    VideoResolutions.Add(resolution);
+                }
+            }
+        }
+        VideoResolutions.Sort();
+        VideoResolutions.Reverse();
```
DaPigGuy, line 119: Since this is the only instance of a list of video resolutions being sorted, the video resolution CompareTo method should probably change so that the default sort order isn't in the reversed order of what we want

NickvisionTubeConverter.Shared/Resources/Strings.resx
```diff
@@ -540,6 +536,10 @@
 		<value>Quality</value>
 	</data>
 
+	<data name="MaxQuality.Field" xml:space="preserve">
+		<value>Maximum quality</value>
```
DaPigGuy, line 540:
```suggestion
		<value>Maximum Quality</value>
```

NickvisionTubeConverter.Shared/Resources/Strings.resx
```diff
@@ -552,8 +552,28 @@
...
 	</data>
 
-	<data name="NewFilename.Field" xml:space="preserve">
-		<value>New Filename</value>
+	<data name="Filename.Field" xml:space="preserve">
+		<value>Filename</value>
+	</data>
+
+	<data name="Crop.Field" xml:space="preserve">
+		<value>Crop thumbnail</value>
```
DaPigGuy, line 560:
```suggestion
		<value>Crop Thumbnail</value>
```

DaPigGuy (changes requested):

DaPigGuy:
Very nice, thanks for picking this up! :)

One big issue; video file types should not be shown as options if the video resolutions list is empty (audio only media). Current behavior causes the quality dropdown to be empty on selecting a video file type, leading to an error on download.

NickvisionTubeConverter.Shared/Models/MediaUrlInfo.cs
fsobolev: > Since this is the only instance of a list of video resolutions being sorted, the video resolution CompareTo method should probably change so that the default sort order isn't in the reversed order of what we want

@nlogozzo Doesn't it interfere with the idea of models being independent of the rest of a program? Making `CompareTo` sorting from greater to smaller can be unexpected if used outside of context of TC.

NickvisionTubeConverter.Shared/Models/VideoResolution.cs
```diff
@@ -0,0 +1,92 @@
...
+    /// </summary>
+    /// <param name="width">The resolution width</param>
+    /// <param name="width">The resolution height</param>
+    public VideoResolution(int width, int height)
+    {
+        Width = width;
+        Height = height;
+    }
+
+    public override string ToString() => $"{Width}x{Height}";
```
nlogozzo, line 31: Add comment
```xml
/// <summary>
/// Gets a string representation of a VideoResolution
/// </summary>
/// <returns>The string representation of the VideoResolution</returns>
```

NickvisionTubeConverter.Shared/Models/MediaUrlInfo.cs
```diff
@@ -0,0 +1,122 @@
...
+                using (Py.GIL())
+                {
+                    outFile.close();
+                }
+                return null;
+            }
+        });
+    }
+
+    private void ParseFromPyDict(PyDict mediaInfo, bool isPartOfPlaylist = false)
```
nlogozzo, line 100: Add Comment

NickvisionTubeConverter.Shared/Models/MediaUrlInfo.cs
```diff
@@ -0,0 +1,122 @@
...
+        {
+            title = title.Replace(c, '_');
+        }
+        foreach (var f in mediaInfo["formats"].As<PyList>())
+        {
+            var format = f.As<PyDict>();
+            if (format.HasKey("vbr"))
+            {
+                var resolution = new VideoResolution(format["width"].As<int>(), format["height"].As<int>());
+                if (!VideoResolutions.Exists(r => r == resolution))
```
nlogozzo, line 113:
```csharp
if(!VideoResolutions.Contains(resolution)) {...}
```
should work...

nlogozzo (changes requested):
Just a quick skim of the code...didn't test functionality or anything yet

NickvisionTubeConverter.Shared/Models/MediaUrlInfo.cs
fsobolev: I tried it first and it didn't work for some reason, so I changed it

nlogozzo: Ah, you have to implement IEquatable<VideoResolution> as well. Look at Transaction model in Denaro. 

After implementation, .Contains should work.

fsobolev:
GUI stuff is done in GNOME for timeframe selection, now I need to figure out how to use ffmpeg downloader with ability to reliably kill it on stop, my old code didn't allow to do that (so I didn't commit it).
Note to myself: https://github.com/yt-dlp/yt-dlp/blob/master/yt_dlp/downloader/external.py#L434

Also, what should we do with WinUI?

nlogozzo:
> Also, what should we do with WinUI?

I'm going to work on converting Application.WinUI backend to use Avalonia instead of WindowsAppSDK. We'll see how easy and quick that goes...

nlogozzo:
> The option can be used with aria2 downloader enabled, it's just that aria2 setting will be ignored. But I'm not sure how to explain this in a good and short way.

If the option is grayed out when speed limit is on, then it's not needed in the description.

I would say for the description just `This option will be ignored if aria2 is used as the downloader.`

fsobolev:
> I would say for the description just `This option will be ignored if aria2 is used as the downloader.`

It's the opposite. `With this option enabled, aria2 downloader can't be used.` or something like this

nlogozzo:
Ah ok, so then `Enabling this option will disable the use of aria2 as the downloader.` should be good.

nlogozzo (approved):

PR #242, opened 2023-05-15
Install yt-dlp via pip and ffmpeg as snap

soumyaDghosh:


snap/snapcraft.yaml
```diff
@@ -95,6 +95,8 @@ parts:
       - cargo
     stage-packages:
       - aria2
+    stage-snaps:
+      - ffmpeg/latest/edge
```
fsobolev, line 99: even in edge ffmpeg is v4.x. I looked at yt-dlp and it turns out there are patched ffmpeg builds from yt-dlp developers that fix some bugs. And according docs, there were additional patches before v5.1 that I believe can affect TC https://github.com/yt-dlp/FFmpeg-Builds#historical-patches
So I guess we better use manual builds until ffmpeg snap gets updated

NickvisionTubeConverter.Shared/Models/Configuration.cs
```diff
@@ -98,6 +98,10 @@ public Configuration()
         {
             Directory.CreateDirectory(ConfigDir);
         }
+        if (!string.IsNullOrEmpty(Environment.GetEnvironmentVariable("SNAP_USER_COMMON")))
+        {
+        PreviousSaveFolder = Environment.GetEnvironmentVariable("SNAP_USER_COMMON");
+        }
         Theme = Theme.System;
         PreviousSaveFolder = Environment.GetFolderPath(Environment.SpecialFolder.MyVideos);
```
fsobolev, line 106: Move your addition below, currently the path will be replaced with `Environment.SpecialFolder.MyVideos` anyway

Issue #29, opened 2022-11-11
Add support for external downloaders


VarLad:
## Problem

While yt-dlp works awesome for Youtube, some sites have different download formats. Many times, downloads from the said websites is really slow. 
yt-dlp supports external downloaders. Taking the aria2 downloader for example, it supports features like multi-threading, cross-platform resuming, download speed limiting, among lots of other features. It works well for other sites as well as external sites. 

How we use it with yt-dlp is:
`yt-dlp --external-downloader aria2c --external-downloader-args "aria2c:-x 16 -k 1M" "Video link here"`

## Feature Request

I've two feature requests for this issue
- Support for external downloaders in yt-dlp
  For this, a drop down menu to select from a list of external downloaders (assuming they're present in user's system) and an empty text box for `--external-downloader-args`.

- Add aria2 binary in the flatpak release. 
  To make it easier for people to use Aria2 external downloader

nlogozzo:
@fsobolev @DaPigGuy maybe we  just use aria2c as the default downloader since it has a lot of nice features over the yt-dlp default?

fsobolev:
I don't mind. Need to investigate how to use an external downloader with new backend.

fsobolev:
https://github.com/scx/uget-flatpak/blob/master/aria2.yaml - example of aria2 in flatpak manifest (we don't need `--enable-bittorrent` but not sure if anything else should be changed)

VarLad:
@nlogozzo Any idea if this could be reopened? 
The original feature request was to implement some options of aria2 as well, like multipart downloading (the number `16` in the command) and minimal split size (1MB in this case) of each part

fsobolev:
> @nlogozzo Any idea if this could be reopened? The original feature request was to implement some options of aria2 as well, like multipart downloading (the number `16` in the command) and minimal split size (1MB in this case) of each part

`-x16` is added [by yt-dlp](https://github.com/yt-dlp/yt-dlp/blob/master/yt_dlp/downloader/external.py#L277]). We can add `-k` as well. Should it be hardcoded or do you think it would be better to allow configuring in settings? 
This is the full command that TC actually runs:

```
[...]

PR #265, opened 2023-05-26
Add Aria2 Configuration Options

### LINKED ISSUE #198: Make aria2 configurable
@fsobolev I believe having the ability to configure both `-x` and `-k` by the user would be really convenient

_Originally posted by @VarLad in https://github.com/NickvisionApps/TubeConverter/issues/29#issuecomment-1514435164_

nlogozzo:
Closes #198


DaPigGuy (approved):

NickvisionTubeConverter.Shared/Models/Download.cs
```diff
@@ -221,9 +223,11 @@ public void Start(bool useAria, bool embedMetadata, string? cookiesPath, Localiz
                     dynamic ariaDict = new PyDict();
                     dynamic ariaParams = new PyList();
                     ariaParams.Append(new PyString($"--max-overall-download-limit={(_limitSpeed ? _speedLimit : 0)}K"));
-                    ariaParams.Append(new PyString("--allow-overwrite=true"));
+                    ariaParams.Append(new PyString($"--allow-overwrite={_overwriteFiles}"));
```
fsobolev, line 226: For aria `allow-overwrite` should always be true to be able to restart downloads.
Also I will improve UI a bit.

fsobolev (changes requested):

NickvisionTubeConverter.Shared/Models/Configuration.cs
```diff
@@ -95,6 +90,8 @@ public Configuration()
         RunInBackground = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
         SpeedLimit = 1024;
         UseAria = false;
+        AriaMaxConnectionsPerServer = 16;
```
nlogozzo, line 93: According to this: https://aria2.github.io/manual/en/html/aria2c.html

The default is 1.

fsobolev: yt-dlp overwrites the default to 16 https://github.com/NickvisionApps/TubeConverter/issues/29#issuecomment-1514422019

fsobolev:
* Added reset buttons to aria settings
* Set -x to be 16 by default, same as yt-dlp's default (https://github.com/NickvisionApps/TubeConverter/issues/29#issuecomment-1514422019)
* Changed MB -> MiB in setting description (according to [aria docs](https://aria2.github.io/manual/en/html/aria2c.html#cmdoption-k), 1M = 1024K, which means it's MiB)

NickvisionTubeConverter.Shared/Models/Configuration.cs
nlogozzo: Ah ok :)

PR #269, opened 2023-05-26
Gnome Circle Tweaks

### LINKED ISSUE #268: GNOME Circle Tweaks

- [x] Filename -> File Name
    - Awaiting clarification
    - Awaiting clarification  
    - image
    -  image

nlogozzo:
Closes #268


NickvisionTubeConverter.Shared/Resources/Strings.resx
```diff
@@ -225,7 +225,7 @@
 	</data>
 
 	<data name="NoDownloads.Description" xml:space="preserve">
-		<value>Add a download to get started.</value>
+		<value>Add a video URL to start downloading.</value>
```
fsobolev, line 228: How about "Add a video, audio or playlist URL to start downloading."?

NickvisionTubeConverter.Shared/Resources/Strings.resx
```diff
@@ -422,7 +422,7 @@
 	</data>
 
 	<data name="NoDownloadsRunning" xml:space="preserve">
-		<value>No downloads running</value>
+		<value>Download Videos</value>
```
fsobolev, line 425: "Download media"?

fsobolev (changes requested):

NickvisionTubeConverter.Shared/Resources/Strings.resx
nlogozzo: That works :)

NickvisionTubeConverter.Shared/Resources/Strings.resx
nlogozzo: Sounds better!

PR #270, opened 2023-05-27
GNOME Circle Tweaks

### LINKED ISSUE #268: GNOME Circle Tweaks

nlogozzo:
Closes #268


nlogozzo:
@fsobolev If you could check the clipboard validation feature :)
PR is pretty much done just waiting for clarification on a few issues on the GitLab post to finish.

NickvisionTubeConverter.GNOME/Views/MainWindow.cs
```diff
@@ -764,4 +786,22 @@ private void RunInBackgroundChanged(object? sender, EventArgs e)
...
         }
     }
+
+    /// <summary>
+    /// Reads the clipboard's text and checks for a valid media url
+    /// </summary>
+    private void ValidateClipboard()
+    {
+        var clipboard = Gdk.Display.GetDefault()!.GetClipboard();
+        _clipboardCallback = async (source, res, data) =>
```
DaPigGuy, line 796: This callback never seems to run, at least when run from GNOME Builder

DaPigGuy: Whoops, it does run but I was printing clipboardText. Real issue seems to be that it doesn't seem to be able to access clipboard contents.

fsobolev: I didn't check anything yet (will do later), but just want to say that I think clipboard validation should happen not only on startup, but every time the main window gets focus.

nlogozzo: > Real issue seems to be that it doesn't seem to be able to access clipboard contents.

Probably need to add a permission for flatpak. Running locally it works fine for me.

nlogozzo: > Or a link can simply be pasted to URL entry when the dialog is opened, I think it's more common pattern (and it's easier to implement I guess).

Yeah I tried doing the validate on focus but couldn't get it to work. This sounds like a good idea to do instead.

fsobolev: For me `clipboardText` is empty too, even when running unsandboxed.

fsobolev: I figured out what was wrong. Reading from clipboard immediately after window creation doesn't give any content, but connecting to `changed` signals shows that the content appears just in a couple of seconds. And I don't quite get how `changed` signal emitted, but it seems to emit every time the window gets focused. I can't say what I like more: a toast with a button or simply automated pasting to URL entry, but what is bad is that currently the link from clipboard is validated by yt-dlp before button to open the dialog appears. If it would be a link to a playlist, it could validate for an eternity. I think the link validation should be just "is it a link or not", and then it will be pasted to the dialog where the actual validation by yt-dlp will get place.

nlogozzo: > I think the link validation should be just "is it a link or not", and then it will be pasted to the dialog where the actual validation by yt-dlp will get place.

You read my mind as I was going to say I want to change it to this

nlogozzo:
Move to #274 with update to gettext

PR #274, opened 2023-05-27
GNOME Circle Tweaks

### LINKED ISSUE #268: GNOME Circle Tweaks

nlogozzo:
Closes #268


nlogozzo:
@fsobolev @DaPigGuy Clipboard updates for you to check:
- Clipboard validation now only checks if the URL scheme is valid instead of actually running it through yt-dlp (yt-dlp validate will occur from the AddDownloadDialog when the user clicks the Validate button like always)
- I can't seem to reproduce your guys issue with not seeing validation on startup. The way I test it is, I copy a like from YouTube and then right away open Tube Converter and it detects it no problem
- Either way if it doesn't work on startup, when the `Add Download` button is clicked, the AddDownloadDialog will check the clipboard again and paste the link into the `Media URL` row if valid:
image

Let me know how it works for you guys :)

fsobolev:
> I can't seem to reproduce your guys issue with not seeing validation on startup. The way I test it is, I copy a like from YouTube and then right away open Tube Converter and it detects it no problem

I guess again like with callbacks its affected by PC speed. For me it needs about a second after a window opens for gdk to get the data from clipboard. If you want to make this feature realible, you could try connecting to `changed` event, and the first time it triggers validate clipboard and then disconnect event.

> when the Add Download button is clicked, the AddDownloadDialog will check the clipboard again and paste the link into the Media URL row if valid:

There's probably nothing we really can do but I dislike how the toast overlaps the button 😞

nlogozzo:
> If you want to make this feature realible, you could try connecting to changed event, and the first time it triggers validate clipboard and then disconnect event.

Ok I'll look it to that. In the mean time if you open AddDownloadDialog it should detect it there and then paste it for you to validate to accommodate that.

> There's probably nothing we really can do but I dislike how the toast overlaps the button 😞

I mean probably not, it's called ToastOverlay so I guess it's expected. We could add a margin to the bottom of the Validate button which will leave room for the overlay. I mean we did get yelled at that in circle review but maybe this time it will be justified 😆

nlogozzo:
> Is there a way to remove downloads? I couldn't find one, at least. Would be nice to have some kind of menu on the rows for that, or selection mode. 
> This menu looks like a save button, but it's something totally different. Given that there's not much in there I think these actions could all just go into the primary menu 

image

Added the option to clear all completed downloads and moved the actions from the `Downloader` menu into buttons in the group headers. I think i'll add text to the buttons too...

nlogozzo:
image

That's better

nlogozzo:
OK, PR ready :)

DaPigGuy:
Having both an icon and label for the buttons seem like an odd choice, especially for the Completed group, where only the `Clear` button has a label. Maybe remove the label?

nlogozzo:
> Having both an icon and label for the buttons seem like an odd choice, especially for the Completed group, where only the Clear button has a label. Maybe remove the label?

I feel like that's pretty common. Here's Settings:
image

As far as only the `Clear` label in `Completed` group, it's the same like in `HeaderBar`s where usually the first button (the most common action) has a label and the others are only icons. Here's Tagger:
image

DaPigGuy:
Fair enough, but definitely deal with the toast getting in the way of the Validate button :sweat_smile:

nlogozzo:
> Fair enough, but definitely deal with the toast getting in the way of the Validate button sweat_smile

Yeah I'm going to try something ...

nlogozzo:
@DaPigGuy Ok try it now., the dialog will grow and shrink for when the toast is shown and dismissed respectively.
The toast will also automatically dismiss once the validate button is pressed (or enter on the url entry).

DaPigGuy:
Seems okay to me

PR #295, opened 2023-05-30
Disallow Conversions

### LINKED ISSUE #263: Prevent conversion
I'd like to use this app to simply download video and audio without any conversions. Lossy conversion degrades quality, and lossless conversion often increases the file size unnecessarily. With this I could finally retire my downloader scripts and start using this app.

There could for example be a new selection in the download window for choosing between video or audio only. Then if best quality is chosen, the file type selection is disabled and no conversion will take place. The best format is automatically chosen.

I know converter is in the name of the app, but I personally don't see the point of converting media files in 2023 unless you want to save disk space.

nlogozzo:
Closes #263


nlogozzo:
@Riksu9000

Ok so here is the option in settings that you can turn on to disable conversions:
image

Once that setting is on, this will be the new AddDownloadDialog:
image

There will simply be two file formats: `Video` and `Audio`. 
- For `Video` you can chose the specific resolution you'd like provided by the native streams 
- For `Audio`, quality is best or worst. 

Then it will just download the format matching that quality directly and won't convert it into any other format or compress it.

NickvisionTubeConverter.Shared/Models/MediaFileType.cs
```diff
@@ -97,6 +103,27 @@ public static class MediaFileTypeHelpers
...
+    /// <returns>True if generic, else false</returns>
+    public static bool GetIsGeneric(this MediaFileType type) => type switch
+    {
+        MediaFileType.MP4 => false,
+        MediaFileType.WEBM => false,
+        MediaFileType.MP3 => false,
+        MediaFileType.M4A => false,
+        MediaFileType.OPUS => false,
+        MediaFileType.FLAC => false,
+        MediaFileType.WAV => false,
```
DaPigGuy, line 124:
```suggestion
```

NickvisionTubeConverter.GNOME/Views/AddDownloadDialog.cs
```diff
@@ -251,10 +267,22 @@ private AddDownloadDialog(Gtk.Builder builder, AddDownloadDialogController contr
...
-            var fileType = (MediaFileType)_fileTypeRow.GetSelected();
-            if (_audioOnly)
+            MediaFileType fileType;
+            if (_controller.DisallowConversions)
+            {
+                fileType = _fileTypeRow.GetSelected() == 0 ? MediaFileType.Video : MediaFileType.Audio;
+                if (_audioOnly)
+                {
+                    fileType = MediaFileType.Audio;
+                }
```
DaPigGuy, line 277:
```suggestion
                fileType = _fileTypeRow.GetSelected() == 0 && !_audioOnly ? MediaFileType.Video : MediaFileType.Audio;
```

DaPigGuy (changes requested):

fsobolev:
I fixed audio extraction.
Also I figured out what `merge_output_format` value is best to use - `null`. Because yt-dlp checks if it's `None`, and I guess when it's used in python it's fine to not set format at all, but in our case it seems it needed to be set like this manually. I tested with different options, works as expected.

PR #309, opened 2023-05-31
fixed snap branch

soumyaDghosh:
merge this asap


fsobolev:
I already restored the branch so snap building is in progress on snapcraft.
Would it make sense to just set source to `.` as you said me?

soumyaDghosh:
no need of that branch, but for now this will be an rc release. And keeping it as source will be fine.

fsobolev (approved):

Issue #173, opened 2023-04-09
Pressing Enter in the 'Add Download' dialog should validate URL, and ESC should "go back"


KSXGitHub:
Since the only purpose of the 'Add Download' dialog is to validate the URL, it makes sense that pressing <kbd>Enter</kbd> in the URL should have the same effect as clicking the 'Validate' button. Pressing <kbd>Enter</kbd> again (after the URL is validated successfully) should start the download process (i.e. same effect as the 'Download' button).

Regarding <kbd>ESC</kbd>, pressing it before the URL is validated should close or hide the 'Add Download' dialog, and pressing it after the URL is validated successfully should restore the 'Add Download' dialog to blank (i.e. same effect as the 'Back' button).

nlogozzo:
https://github.com/nlogozzo/NickvisionTubeConverter/releases/tag/2023.4.1

PR #317, opened 2023-06-02
GNOME Circle

DaPigGuy:
https://gitlab.gnome.org/Teams/Circle/-/issues/128#note_1758431

- [ ] The summary "Get video and audio from the web" could be shorter/catchier. How about something like "Download web videos"?
- [x] The terminal output should be cut off at the bottom edge of the card, rather than 5px above it
 - [x] The divider line between the top part of the card and the terminal output should go edge to edge, spanning the entire width of the card

### DIFF
--- a/NickvisionTubeConverter.GNOME/Blueprints/download_row.blp
+++ b/NickvisionTubeConverter.GNOME/Blueprints/download_row.blp
@@ -4,13 +4,12 @@ using Adw 1;
 Adw.Bin _root {
   Gtk.Box {
     orientation: vertical;
-    margin-start: 6;
-    margin-top: 6;
-    margin-end: 6;
-    margin-bottom: 6;
 
     Gtk.Box {
       orientation: horizontal;
+      margin-start: 6;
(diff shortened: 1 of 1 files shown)


DaPigGuy:
There's currently a big accessibility issue with the Add Download dialog where pressing the Enter key triggers the default widget (Download button) despite being focused on a combo row

nlogozzo:
> There's currently a big accessibility issue with the Add Download dialog where pressing the Enter key triggers the default widget (Download button) despite being focused on a combo row

That's a result of #173

DaPigGuy:
> > There's currently a big accessibility issue with the Add Download dialog where pressing the Enter key triggers the default widget (Download button) despite being focused on a combo row
> 
> That's a result of #173

Well it broke keyboard navigation on that dialog as a result :sweat_smile:

nlogozzo:
Merging this now for V2023.6.0 as it doesn't require any translation updates...will not delete branch so we can finish the other stuff for V2023.6.1 :)

PR #325, opened 2023-06-05
Description + Username/Password Fields + Remember NumberTitles

### LINKED ISSUE #319: GNOME Circle Tweaks
- [x] The summary "Get video and audio from the web" could be shorter/catchier. How about something like "Download web videos"?

### LINKED ISSUE #322: Add support for yt-dlp usernames and passwords
yt-dlp provides support for the `--username` and `--password` tags in order to download from a website that requires an account. The recent addition of the cookie support works for some sites, but in cases such as https://nebula.tv/ a passed username and password is required to download

### LINKED ISSUE #324: Add "Include Number Tiltle for Playlists" default in Preferences
While it is possible to do this when selecting the download, activating this behaviour needs to be done each time. 

I was wondering if it would be possible to add "Include Number Titles when downloading Playlists" in Preferences so the user can set this behaviour as default?

Alternatively, it could be remembered as default like it's done for the Download folder location?

nlogozzo:
Closes #319 
Closes #322 
Closes #324 

## TODO
- [x] Update banner (@fsobolev could you take care of this?)


fsobolev:
I didn't test it yet, will do a bit later, but I was the screenshot in the issue.
I don't think "Number Titles" should be on main view, because it affects titles that are shown on a separate view. Perhaps we could have 2 synced switches on both views?

nlogozzo:
> I don't think "Number Titles" should be on main view, because it affects titles that are shown on a separate view. Perhaps we could have 2 synced switches on both views?

I mean we *could* have two synced switches, but I don't think that's ideal.
I think the switch should be moved to the main view as for many users who just want to download a playlist without editing the titles individually, they can just turn on this switch without going into the sub-page.

On second thought however, maybe instead of having two switches synced, we can bring back the toggle button on the sub-page and sync that with the main switch.

fsobolev:
> I mean we _could_ have two synced switches, but I don't think that's ideal. I think the switch should be moved to the main view as for many users who just want to download a playlist without editing the titles individually, they can just turn on this switch without going into the sub-page.
> 
> On second thought however, maybe instead of having two switches synced, we can bring back the toggle button on the sub-page and sync that with the main switch.

I see the point of making the switch easier to access by moving it to the main page, but at the same time it affects things on advanced page, so it makes sense to also have it there. And I don't get what's the benefit of using toggle button instead of switch.

fsobolev:
> I just thought the ToggleButton looks better on that page like it is now

I think using a row with switch in both places is less confusing, option is the same and so appearance of controls to change it is the same.

nlogozzo:
> I think using a row with switch in both places is less confusing, option is the same and so appearance of controls to change it is the same.

Ok fixed with https://github.com/NickvisionApps/TubeConverter/pull/325/commits/a451dd34c70dd1d2b38e1e36895ff4291ac499c5

nlogozzo:
Once @fsobolev updates the banner, i'd say this PR is ready. 

Keyring will be implemented in TC later on, once [Proposals/#6](https://github.com/NickvisionApps/Proposals/issues/6 ) is finished. Therefore, it shouldn't block this PR.

fsobolev:
Gonna update the banners and then I think it's good to go

PR #349, opened 2023-06-11
Auto Retry Once

### LINKED ISSUE #343: Retry failed downloads
Hi, me again, I was wondering if it would be possible for TubeConverter to retry up to X many times failed downloads, with X being set in preferrences (perhaps with a default like 2 and max maxum 10).

It's pretty ocmmon (for me ) to have downloads fail but when I try again, they work. it would be ideal if this was done automatically.

nlogozzo:
Closes #343 

This will automatically retry a download once. If it fails after that one automatic retry, it will be up to the user to manually retry it. 

If a download was stopped, it will not be automatically retried. Only if it completed and failed.


fsobolev:
Will a notification about failed download be only shown after retry?

nlogozzo:
> Will a notification about failed download be only shown after retry?

No, it will be shown before as 
```
DownloadCompleted?.Invoke(this, (download.Id, successful, download.Filename));
``` 
is still invoked.

So let's say a download fails and then auto retries and is successful, it will show a notification that it failed and then that it was successful.

fsobolev:
I think a notification should only be shown after retry. There's no action expected from user on auto retry and so nothing to notify about.

nlogozzo:
> I think a notification should only be shown after retry. There's no action expected from user on auto retry and so nothing to notify about.

Yeah your right, let me fix that.

nlogozzo:
Ok, done!

fsobolev (approved):

PR #352, opened 2023-06-11
Specific Timeframe

### LINKED ISSUE #105: Download only specified timeframe

nlogozzo:
Closes #105 

I also cleaned up the code a bunch, moving most logic from the UI to controller as it should be. Please test and make sure nothing broke.


nlogozzo:
Ok @fsobolev I got stopping the downloads when using specific timeframe to work. I use a new python module called `psutil` that returns a set of child processes belonging to the download's python process. So I basically iterate through all of them and check which ones are called `ffmpeg` and kill them. And it works! I added the module to flatpak manifest, but if testing locally make sure you run `pip install psutil` before...

Also, I rewrote a lot of the code in general, moving things from UI side to backend side as it should be, so please double test everything and make sure it's all good :)

nlogozzo:
I'm having issues with the flatpak: https://github.com/NickvisionApps/TubeConverter/actions/runs/5239738030/jobs/9459853998

From the log, you can see that `wheel` says it's missing `flit-core`, but right above it, `flit-core` is built successfully. Upon google searching, it seems like:
https://stackoverflow.com/questions/76357177/pip-cant-find-flit-core-dependency-when-offline-installing-other-packages
and
https://github.com/pypa/flit/issues/363
but without these dependencies we can't install psutil....so might need to find another way to implement killing ffmpeg if this doesn't work. 

*Can we maybe use a [venv](https://docs.python.org/3/tutorial/venv.html) inside flatpak build? IDK*

nlogozzo:
> I'm having issues with the flatpak: https://github.com/NickvisionApps/TubeConverter/actions/runs/5239738030/jobs/9459853998

Fixed in latest commit. Got `psutil` working for flatpak and snap!

NickvisionTubeConverter.Shared/Models/Timeframe.cs
```diff
@@ -0,0 +1,67 @@
...
+        }
+        catch
+        {
+            throw new ArgumentException("Unable to parse emd time.");
+        }
+        if(startSpan < TimeSpan.FromSeconds(0))
+        {
+            throw new ArgumentException("Invalid start time.");
+        }
+        if(endSpan < startSpan + TimeSpan.FromSeconds(1) || endSpan > TimeSpan.FromSeconds(duration))
```
DaPigGuy, line 61: Why +1 second?

NickvisionTubeConverter.Shared/Controllers/AddDownloadDialogController.cs
```diff
@@ -110,88 +152,123 @@ public async Task<List<string>> GetKeyringCredentialNamesAsync()
...
-    public void ToggleNumberTitles(MediaUrlInfo mediaUrlInfo, bool toggled)
+    /// <param name="toggled">Whether or not to number titles</param>
+    public void ToggleNumberTitles(bool toggled)
     {
         var numberedRegex = new Regex(@"[0-9]+ - ", RegexOptions.None);
-        for (var i = 0; i < mediaUrlInfo.MediaList.Count; i++)
+        for (var i = 0; i < _mediaUrlInfo.MediaList.Count; i++)
         {
-            if (toggled)
+            if(_mediaUrlInfo != null)
```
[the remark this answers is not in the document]
nlogozzo: Your right, that if statement should be moved before the for loop.

NickvisionTubeConverter.Shared/Models/Timeframe.cs
nlogozzo: That's a question for @fsobolev,  this was his code from an old branch we had.

NickvisionTubeConverter.Shared/Controllers/AddDownloadDialogController.cs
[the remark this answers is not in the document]
nlogozzo: https://github.com/NickvisionApps/TubeConverter/pull/352/commits/3afac47d927db35c0c0e2d944c25ecd430316b31

NickvisionTubeConverter.Shared/Models/Timeframe.cs
nlogozzo: @DaPigGuy Ah i see why he did that.

Let say you have a start time of `00:00:00`. Without adding `TimeSpan.FromSeconds(1)`, `00:00:00` will be considered a valid end time. But it's not because your downloading nothing. Therefore `00:00:01` and above will be considered valid.

DaPigGuy: Wouldn't a `<=` comparison be more appropriate

DaPigGuy (approved):
See the one unresolved comment

nlogozzo:
> See the one unresolved comment

Yes that would work too...I'll change it then merge. Thanks :)

PR #408, opened 2023-06-26
TubeConverter -> Parabolic (links and docs)

fsobolev:


nlogozzo (changes requested):
Can you rename `TubeConverter.doap` to `Parabolic.doap` since that it now the name of the repo?

nlogozzo (approved):

PR #419, opened 2023-06-29
using ffmpeg content snap instead of building it from source inside the snap

soumyaDghosh:


soumyaDghosh:
Also, renamed the artifact file in workflow, cause it's the file which the snapcraft produces in real.

nlogozzo:
> Also, renamed the artifact file in workflow, cause it's the file which the snapcraft produces in real.

That's fine...however it seems the `snap` branch is out of date with `main`. We no longer have `build.yml` but a separate `snap.yml` to build `flatpak` and `snap` in separate CIs.

soumyaDghosh:
Too much hell to resolve this conflicts, so reopening this PR

PR #421, opened 2023-06-29
Enhancements to the snap

soumyaDghosh:


nlogozzo (approved):

.github/workflows/snap.yml
```diff
@@ -1,8 +1,8 @@
 on:
   push:
-    branches: [ "snap" ]
+    branches: [ "main" ]
   pull_request:
-    branches: [ "snap" ]
+    branches: [ "main" ]
```
fsobolev, line 5: why?

soumyaDghosh: No need for a separate snap branch, it seems useless as there's nothing to patch on. Keeping it simple. A new update, change the source-tag

PR #433, opened 2023-07-04
Decrease CPU usage when downloading

### LINKED ISSUE #432: High CPU usage and unresponsiveness when downloading
Platform: i5, Gnome 44, Wayland, openSUSE Tumbleweed, Flatpak v2023.7.0

I'm seeing high CPU usage according to htop (around 116%) for the flatpak app whenever it is downloading. Even with just a single download stream from twitch. 
The CPU usage drops whenever the download slows down or halts for a second, but rises when the download speed picks up again. Using yt-dlp on the terminal for the same file yields almost no CPU usage. 
Also, if the stream is rather long, 10h or so, the whole window becomes unresponsive and requires force-quitting before the download is complete.
Yt-dlp seems to be producing quite a bit of log entries, updating the progress indicator multiple times per second. Could log spam be the issue?

fsobolev:
- `noprogress` option added to yt-dlp, so it doesn't show download progress in the log (the progress is still shown in the downloads row)
- TextView only gets updated when the log changes

When downloading [this video](https://www.youtube.com/watch?v=zrThhl9tR5E&t=1s), CPU usage dropped from 25-30% to 4-6% on my Celeron N5095.
Closes #432.

Other changes:
- Updated cake building in flatpak like in Cavalier
- Forced `en_US` locale for Fleet run config, because for some reason it was running with `POSIX` locale for me, causes Parabolic to not download anything because it seems python expects UTF8 locale to properly work with strings.


nlogozzo (approved):

nlogozzo:
Thank you :) feel free to release an update when you have time!

PR #435, opened 2023-07-05
Write playlist position to metadata track number

### LINKED ISSUE #417: Add "Write Track Number Metadata" Option To Playlist Page

fsobolev:
Closes #417

I initially implemented it with a switch in add download dialog to disable it, but then I thought: what is the use-case for disabling it and is there any harm in writing track tag also for video? I think no use-case and no harm, so it just writes track number whenever the container allows it (MP4 doesn't).

When downloading single media, 0 will be written, but yt-dlp already did this by default (at least for audio files), so there's no changes here.

### DIFF
--- a/NickvisionTubeConverter.Shared/Models/MediaUrlInfo.cs
+++ b/NickvisionTubeConverter.Shared/Models/MediaUrlInfo.cs
@@ -83,18 +83,20 @@ public class MediaUrlInfo
                     }
                     if (mediaInfo.HasKey("entries"))
                     {
+                        var i = 1u;
                         foreach (var e in mediaInfo["entries"].As<PyList>())
                         {
                             if (e.IsNone())
                             {
                                 continue;
                             }
-                            mediaUrlInfo.ParseFromPyDict(yt, e.As<PyDict>(), true, url);
+                            mediaUrlInfo.ParseFromPyDict(yt, e.As<PyDict>(), i, url);
+                            i++;
                         }
                     }
                     else
                     {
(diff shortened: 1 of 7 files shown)


NickvisionTubeConverter.Shared/Models/MediaUrlInfo.cs
```diff
@@ -83,18 +83,20 @@ private MediaUrlInfo(string url)
...
+                        var i = 1u;
                         foreach (var e in mediaInfo["entries"].As<PyList>())
                         {
                             if (e.IsNone())
                             {
                                 continue;
                             }
-                            mediaUrlInfo.ParseFromPyDict(yt, e.As<PyDict>(), true, url);
+                            mediaUrlInfo.ParseFromPyDict(yt, e.As<PyDict>(), i, url);
+                            i++;
```
fsobolev, line 94: I'm not entirely sure about this, so let me explain. Let's say we have 3 items playlist with the 2nd one being unavailabe. With this code, available items will be numbered 1 and 2 (in track metadata and when numbering titles), this is the old behaviour and that's why I kept it.
But I can change it so numbers will represent the actual position of media in playlist, so in this example items will be numbered 1 and 3.
I'm not sure what is better.

nlogozzo (approved):

nlogozzo (changes requested):
For changelog:
- No periods
- Make it: "When downloading a playlist, a media's position will be written to the tag's track property"

nlogozzo (approved):

PR #468, opened 2023-07-21
Download History

### LINKED ISSUE #453: [Feature Request] Searchable download history
It'd be nice to store download history and make it searchable, so we can quickly search and open previously downloaded videos.

(Bonus points for keeping it in a documented format or sqlite file :smile: that can be used from other apps/scripts).

I currently store all downloaded videos in a local folder, so I can use the CLI or Nautilus to search and open them, but this currently lacks advanced search capabilities than an integrated solution storing video metadata would give us, I think.

Props for the amazing app :sparkles:

nlogozzo:
Closes #453 

image

Download history is saved in `downloadHistory.json` in the config directory of the application (same location as `config.json`)


NickvisionTubeConverter.GNOME/Controls/HistoryDialog.cs
```diff
@@ -0,0 +1,127 @@
...
+            if(string.IsNullOrEmpty(pair.Value))
+            {
+                row.SetTitle(pair.Key);
+            }
+            else
+            {
+                row.SetTitle(pair.Value);
+                row.SetSubtitle(pair.Key);
+            }
+            row.SetTitleLines(1);
```
fsobolev, line 61: `SetSubtitleLines(1)` too

NickvisionTubeConverter.GNOME/Views/AddDownloadDialog.cs
```diff
@@ -306,6 +306,7 @@ public async Task PresentAsync(string? url = null)
         //Validated from startup
         if (!string.IsNullOrEmpty(url))
         {
+            _urlRow.SetText(url);
             await SearchUrlAsync(url);
```
fsobolev, line 310: So it will paste the link and automatically start validation, correct? Maybe just pasting is enough?

fsobolev (changes requested):

NickvisionTubeConverter.GNOME/Views/AddDownloadDialog.cs
nlogozzo: Yes it will automatically paste and start validation....if the user clicks the button to start a media again, then i'm 99% sure they'll press the validation button to configure the download.

It's different than reading from clipboard because the user might have had an old url or something in the clipboard that is different than what they'd like to download. This is more likely to occur hence a clipboard link isn't automatically validated.

fsobolev (approved):

nlogozzo:
Sorry, changed it quickly where the date of the download will be stored as well that way we can sort the history with most recent downloads at top.

PR #471, opened 2023-07-22
Support Opening History Downloads + Empty History State

nlogozzo:
As per #453:
> so we can quickly search and open previously downloaded videos.

This PR makes it where the history will also save the path of a downloaded media. If that path still exists when the HistoryDialog is opened, an open button will be added to the row to allow the user to open the folder of that downloaded media:

image

### DIFF
--- a/NickvisionTubeConverter.GNOME/Controls/HistoryDialog.cs
+++ b/NickvisionTubeConverter.GNOME/Controls/HistoryDialog.cs
@@ -84,7 +84,7 @@ public partial class HistoryDialog : Adw.Window
             {
                 var openButton = Gtk.Button.New();
                 openButton.SetIconName("document-open-symbolic");
-                openButton.SetTooltipText(_("Open in File Manager"));
+                openButton.SetTooltipText(_("Open"));
                 openButton.SetValign(Gtk.Align.Center);
                 openButton.AddCssClass("flat");
                 openButton.OnClicked += (sender, e) =>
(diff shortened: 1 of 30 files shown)


NickvisionTubeConverter.GNOME/Controls/HistoryDialog.cs
```diff
@@ -59,17 +70,31 @@ private HistoryDialog(Gtk.Builder builder, Gtk.Window parent, string iconName, D
...
                 DownloadAgainRequested?.Invoke(this, pair.Key);
             };
-            row.AddSuffix(button);
-            row.SetActivatableWidget(button);
+            row.AddSuffix(downloadButton);
+            if(File.Exists(pair.Value.Path))
+            {
+                var openButton = Gtk.Button.New();
+                openButton.SetIconName("document-open-symbolic");
+                openButton.SetTooltipText(_("Open"));
```
fsobolev, line 87: Does it open file or folder with file selected? If the former, then "play" icon would be better. If the latter, I would change tooltip to "Open in Folder".

fsobolev (changes requested):

fsobolev:
Also I think buttons should be in the opposite order: open first, then download again, so download again buttons will be in one column.

nlogozzo:
Also added an empty history state:
image

NickvisionTubeConverter.GNOME/Controls/HistoryDialog.cs
nlogozzo: it's the former so I will change to play icon

fsobolev (approved):

PR #481, opened 2023-07-25
Fix invalid video resolution options

DaPigGuy:
Fixed non-video formats (thumbnails?) resolutions being included in the video resolution options, resulting in errors when attempting to download.


nlogozzo:
Not on my PC right now, so can you confirm that this video still validates:

https://www.zdf.de/show/mai-think-x-die-show/maithink-x-folge-23-100.html

DaPigGuy:
Yes, it still validates. Funny enough, I've already tested that URL after `git blame`ing and finding your commit.

fsobolev:
I confirm that various media files validate and download without issues, and there's no format options with exteremely low resolutions that was causing errors.

fsobolev (approved):

PR #486, opened 2023-07-28
GNOME - Select/deselect all in playlist

### LINKED ISSUE #478: [Feature Request] Ability to select all or deselect all when downloading playlists
Given that I update the videos I download frequently, I need to be able to download only a specific number of videos in playlists, but I can't deselect everything in bulk. Alternatively, I also can't choose to select all. I would just like to ask if a feature can be implemented that would be able to address these, thanks!

fsobolev:
Closes #478 

Снимок экрана от 2023-07-28 23-32-07


DaPigGuy (approved):
Code LGTM and works from my quick testing w/ a playlist

--- 

Edit: you can click Download with 0 playlist videos selected, but that's not a new issue

NickvisionTubeConverter.GNOME/Controls/MediaRow.cs
```diff
@@ -15,6 +15,11 @@ public class MediaRow : Adw.EntryRow
     [Gtk.Connect] private readonly Gtk.CheckButton _downloadCheck;
     [Gtk.Connect] private readonly Gtk.Button _undoButton;
 
+    public bool CheckActive
```
nlogozzo, line 18: I would just call the property `Active` and add a get for it too

```csharp
/// <summary>
/// The active status of the row's check button
/// </summary>
public bool Active
{
    get => _downloadCheck.GetActive();

    set => _downloadCheck.SetActive(value);
}
```

nlogozzo (changes requested):
If you could also change the version to `2023.8.0-next` and update the changelog :)
And `GeneratePot` and `UpdatePo` as well

nlogozzo (approved):

nlogozzo:
Thanks :)

PR #491, opened 2023-07-29
Test Internet Connection On Startup

### LINKED ISSUE #476: Detect Network Connectivity
Add a feature where Parabolic detects network connectivity and informs the user if no network is available.

nlogozzo:
Closes #476 

image

### DIFF
--- a/NickvisionTubeConverter.Shared/Controllers/MainWindowController.cs
+++ b/NickvisionTubeConverter.Shared/Controllers/MainWindowController.cs
@@ -7,6 +7,7 @@ using Python.Runtime;
 using System;
 using System.IO;
 using System.Net.NetworkInformation;
+using System.Net.Sockets;
 using System.Threading.Tasks;
 using static NickvisionTubeConverter.Shared.Helpers.Gettext;
 
@@ -181,6 +182,17 @@ public class MainWindowController : IDisposable
         {
             NotificationSent?.Invoke(this, new NotificationSentEventArgs(_("No active internet connection"), NotificationSeverity.Error, "no-network"));
         }
+        NetworkChange.NetworkAvailabilityChanged += async (sender, e) =>
+        {
+            if (await CheckNetworkConnectivityAsync())
+            {
+                NotificationSent?.Invoke(this, new NotificationSentEventArgs("", NotificationSeverity.Success, "network-restored"));
+            }
+            else
+            {
+                NotificationSent?.Invoke(this, new NotificationSentEventArgs(_("No active internet connection"), NotificationSeverity.Error, "no-network"));
+            }
+        };
     }
 
     /// <summary>
@@ -189,9 +201,23 @@ public class MainWindowController : IDisposable
     /// <returns>True if network connection active, else false</returns>
     public async Task<bool> CheckNetworkConnectivityAsync()
     {
-        using var ping = new Ping();
-        var reply = await ping.SendPingAsync("8.8.8.8"); //google
-        return reply.Status == IPStatus.Success;
+        foreach (var addr in new[] { "8.8.8.8", "http://www.baidu.com", "http://www.aparat.com" })
+        {
+            try
+            {
+                using var ping = new Ping();
+                var reply = await ping.SendPingAsync(addr);
+                if (reply.Status == IPStatus.Success)
+                {
+                    return true;
+                }
+            }
+            catch
+            {
+                return false;
+            }
+        }
+        return false;
     }
 
     /// <summary>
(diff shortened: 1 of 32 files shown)


NickvisionTubeConverter.Shared/Controllers/MainWindowController.cs
```diff
@@ -172,6 +176,22 @@ public async Task StartupAsync()
...
+    }
+
+    /// <summary>
+    /// Checks for an active network connection
+    /// </summary>
+    /// <returns>True if network connection active, else false</returns>
+    public async Task<bool> CheckNetworkConnectivityAsync()
+    {
+        using var ping = new Ping();
+        var reply = await ping.SendPingAsync("8.8.8.8"); //google
```
fsobolev, line 193: I wonder how well this is going to work where Google services are fully or partially blocked.
There's [NetworkAvailabilityChanged](https://learn.microsoft.com/en-us/dotnet/api/system.net.networkinformation.networkchange.networkavailabilitychanged?view=net-7.0#system-net-networkinformation-networkchange-networkavailabilitychanged) event that we could probably use and show banner not only on start.

nlogozzo: The problem is i believe NetworkAvaliablity APIs only tell if your connected to a network, not an active internet connect. So you can be connected to your router's network (LAN) but have no internet connection to the outside world. So it'll say you have a network connection, but you can't access the internet.

So we need a reliable site (could be literally anything) on the internet that we could ping to check if we have active internet. We could even use nickvision.org lol . I just wanted to use something more robust because if said site goes down, it'll say you have no internet.

fsobolev: Then I have a different idea: we could ask NetworkManager using Tmds.Dbus. I can try to implement this tonight.

nlogozzo: The thing is I would like to keep it cross platform for android app and eventually windows...maybe we can just ping gnome.org would work?

DaPigGuy: [Gio.NetworkMonitor](https://docs.gtk.org/gio/method.NetworkMonitor.get_connectivity.html)

nlogozzo: We're going to go with this solution: https://stackoverflow.com/questions/2031824/what-is-the-best-way-to-check-for-internet-connectivity-using-net/2031831#2031831

To keep windows/android support in backend

DaPigGuy: Well you can still use the NetworkMonitor `network-changed` signal on the GNOME implementation to automatically recheck

nlogozzo: > Well you can still use the NetworkMonitor `network-changed` signal on the GNOME implementation to automatically recheck post-startup

That would work :)

fsobolev: I'm not sure splitting network monitoring in 2 assemblies is a good idea. We could do that instead: if there's at least one active download in downloading state and current total speed is 0, check network using pings every N seconds (5? 10?).

nlogozzo: > I'm not sure splitting network monitoring in 2 assemblies is a good idea. We could do that instead: if there's at least one active download in downloading state and current total speed is 0, check network using pings every N seconds (5? 10?).

That could work too...but i just remembered csharp's networkinformation itself also provides a network changed event that we can subscribe too and ping to see if it's an active internet connection.

NickvisionTubeConverter.Shared/Controllers/MainWindowController.cs
```diff
@@ -172,6 +177,44 @@ public async Task StartupAsync()
...
+        };
+    }
+
+    /// <summary>
+    /// Checks for an active network connection
+    /// </summary>
+    /// <returns>True if network connection active, else false</returns>
+    public async Task<bool> CheckNetworkConnectivityAsync()
+    {
+        foreach (var addr in new[] { "8.8.8.8", "http://www.baidu.com", "http://www.aparat.com" })
```
nlogozzo, line 204: Could you make the array a private readonly variable, set in the constructor, and used here? That way we avoid constantly making a new array allocation everytime we check network...

PR #494, opened 2023-07-30
Use Aura

fsobolev:


NickvisionTubeConverter.Shared/Models/DownloadHistoryItem.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+
+namespace NickvisionTubeConverter.Shared.Models;
+
+/// <summary>
+/// Item in download history
+/// </summary>
+public class DownloadHistoryItem
+{
```
nlogozzo, line 9: We should have a `public string Url { get; set; }` property and a Constructor such as this
```csharp
public DownloadHistoryItem(string url)
{
    Url = url;
    Title = "";
    Date = DateTime.Now;
    Path = "";
}
```

NickvisionTubeConverter.Shared/Models/DownloadManager.cs
```diff
@@ -200,9 +201,14 @@ public void AddDownload(Download download, DownloadOptions options)
             _queued.Add(download.Id, (download, options));
             DownloadAdded?.Invoke(this, (download.Id, download.Filename, download.SaveFolder, false));
         }
-        var history = DownloadHistory.Current;
-        history.History[download.MediaUrl] = (Path.GetFileNameWithoutExtension(download.Filename), DateTime.Now, $"{download.SaveFolder}{Path.DirectorySeparatorChar}{download.Filename}");
-        history.Save();
+        var history = (DownloadHistory)Aura.Active.ConfigFiles["downloadHistory"];
+        history.History[download.MediaUrl] = new DownloadHistoryItem
```
nlogozzo, line 205: Using the new constructor this will be
```csharp
history.History[download.MediaUrl] = new DownloadHistoryItem(download.MediaUrl)
{
    Title = Path.GetFileNameWithoutExtension(download.Filename),
    Path = $"{download.SaveFolder}{Path.DirectorySeparatorChar}{download.Filename}"
};
```

nlogozzo (changes requested):

NickvisionTubeConverter.Shared/Models/DownloadHistoryItem.cs
fsobolev: what's the point of keeping URL in HistoryItem if it's already used as the key in history dictionary?

nlogozzo: Because for the sake of object design, this object doesn't necessarily need to be used in a dictionary. Therefore the Url property will become the only way to differentiate the objects.

nlogozzo: and we can even clean up the config file and just use `List<DownloadHistoryItem>` but the dictionary ensures we have a single `DownloadHistoryItem` per Url. I might be able to change it into a `Set<DownloadHistoryItem>` will look into it.

NickvisionTubeConverter.Shared/Controllers/MainWindowController.cs
```diff
@@ -180,13 +212,10 @@ public async Task StartupAsync()
...
-        if (!await CheckNetworkConnectivityAsync())
+        _netmon = await NetworkMonitor.NewAsync();
+        _netmon.StateChanged += (sender, state) =>
         {
-            NotificationSent?.Invoke(this, new NotificationSentEventArgs(_("No active internet connection"), NotificationSeverity.Error, "no-network"));
-        }
-        NetworkChange.NetworkAvailabilityChanged += async (sender, e) =>
-        {
-            if (await CheckNetworkConnectivityAsync())
+            if (state == NetworkState.ConnectedGlobal)
```
DaPigGuy, line 218: The `no-network` event is never sent for the banner

NickvisionTubeConverter.Shared/Models/DownloadHistoryItem.cs
nlogozzo: > and we can even clean up the config file and just use List<DownloadHistoryItem> but the dictionary ensures we have a single DownloadHistoryItem per Url. I might be able to change it into a Set<DownloadHistoryItem> will look into it.

seems like Dictionary is the best structure to use for the case of keeping one recent history item per url. But please still add the Url property and new constructor for API design...

similar to how in Denaro we have

```csharp
class Transaction 
{
     public uint ID { get; init; }
     
     public Transaction(uint id) { ... }
}

public Dictionary <uint, Transaction> Transactions ...
```
The only thing is for `DownloadHistoryItem` `Url` needs to have `set` for JsonSerialization and Deserialization to work.

nlogozzo (approved):

PR #496, opened 2023-07-30
Proxy Preference

### LINKED ISSUE #485: Ability to select a proxy
yt-dlp through its `--proxy` CLI option can accept a variety of protocols for proxying the connection.  While it is possible to set the proxy from GNOME settings (and expect it would work, although frankly only `https_proxy` worked from me testing this app), this is sadly not the case for isolated environments such as Flatpak.
For those you will have to set the environment variable (`https_proxy`/`all_proxy`) manually when launching the app from CLI.

An option in the app settings to set a proxy should make this easier.

Perhaps, it could be interesting to see something like "Pass additional flags to yt-dlp" option, although not sure how possible is this.

nlogozzo:
Closes #485


fsobolev (approved):

NickvisionTubeConverter.Shared/Models/DownloadOptions.cs
```diff
@@ -43,20 +47,22 @@ public class DownloadOptions
...
-    public DownloadOptions(bool overwriteExistingFiles, bool useAria, string? cookiesPath, int ariaMaxConnectionsPerServer, int ariaMinSplitSize, string subtitleLangs, bool embedMetadata, bool embedChapters)
+    public DownloadOptions(bool overwriteExistingFiles, bool useAria, int ariaMaxConnectionsPerServer, int ariaMinSplitSize, string subtitleLangs, string proxyUrl, string cookiesPath, bool embedMetadata, bool embedChapters)
     {
         OverwriteExistingFiles = overwriteExistingFiles;
         UseAria = useAria;
-        CookiesPath = cookiesPath;
         AriaMaxConnectionsPerServer = ariaMaxConnectionsPerServer;
         AriaMinSplitSize = ariaMinSplitSize;
         SubtitleLangs = subtitleLangs;
+        ProxyUrl = "";
```
fsobolev, line 64: I was blind, sorry @nlogozzo

nlogozzo: Ah my fault!! Will fix 😆

PR #522, opened 2023-08-05
Shared - Prefer non vp* codec when downloading mp4

### LINKED ISSUE #520: Allow codec choice for Youtube downloads
It's possible to pick a video format (either webp or mp4) for Youtube downloads, but both encapsule VP9 which is not supported on many TVs and multimedia boxes, that's why it'd be great to be able to pick H.264 codec in which Youtube videos are also provided.

fsobolev:
Closes #520


nlogozzo (approved):

nlogozzo:
Will release next beta when I get home :)

PR #570, opened 2023-09-06
Updated Aura with secret service support

### LINKED ISSUE #569: Update Aura to 2023.9.1

fsobolev:
Closes #569
Depends on https://github.com/NickvisionApps/Aura/pull/18
Depends in https://github.com/NickvisionApps/Aura/pull/23

TODO:
- [x] Update csproj to reference new version of Aura
- [x] Fix creating Parabolic keyring when keyring is locked
- [x] Update po(t)


fsobolev:
Forgot to update flatpak sources, and also need to check whether flatpak requires any new permissions, will build locally to test

nlogozzo:
> Forgot to update flatpak sources, and also need to check whether flatpak requires any new permissions, will build locally to test

Probably need `"--talk-name=org.freedesktop.secrets"`

flatpak/org.nickvision.tubeconverter.json
```diff
@@ -171,6 +172,7 @@
                 }
             ]
         },
+        "libsecret/libsecret.json",
```
nlogozzo, line 175: Why couldn't you just put the libsecret module here....no need to keep it in another file

nlogozzo (changes requested):
Another question: What happens to Keyrings created with old Parabolic and then opened in this new Parabolic? Will they still be asked for the password everytime to open the Keyring? Or the first time they open the Keyring it will get stored in libsecret and they won't be asked again?

flatpak/org.nickvision.tubeconverter.json
fsobolev: Indeed, sorry, I first used libsecret from shared-modules and so placed it in separate folder (the same as with intltool), I will move it.

fsobolev:
> Another question: What happens to Keyrings created with old Parabolic and then opened in this new Parabolic? Will they still be asked for the password everytime to open the Keyring? Or the first time they open the Keyring it will get stored in libsecret and they won't be asked again?

I didn't test it myself, but I expect that a new random password will be generated and saved in system keyring (only at first start) but because it will not match the keyring's password the user will be prompted every time. There are 2 ways to migrate: either remove keyring and create new one, or save the keyring passord to system keyring, this can be easily done using Seahorse, I want to write a small guide as soon as the beta releases.

flatpak/org.nickvision.tubeconverter.json
```diff
@@ -172,7 +172,7 @@
                 }
             ]
         },
-        "libsecret/libsecret.json",
+        "libsecret.json",
```
nlogozzo, line 175: What I meant it why have a separate `libsecret.json` file all together? Just paste it inside the manifest itself...

nlogozzo (changes requested):

flatpak/org.nickvision.tubeconverter.json
fsobolev: Ah, okay

nlogozzo:
Button to Docs in PasswordDialog:
image

nlogozzo (approved):

nlogozzo:
@soumyaDghosh is there anything that needs to be added to the snap yml for `libsecret` use?

soumyaDghosh:
It's in gnome-sdk

fsobolev:
> It's in gnome-sdk

Please check manually that it works (simply enabling keyring is enough), there's some weird thing in sandboxed environment so we have to build libsecret with `-Dgcrypt=false` even tho libsecret is in GNOME Platform in flatpak too.

fsobolev:
> Please check manually that it works (simply enabling keyring is enough), there's some weird thing in sandboxed environment so we have to build libsecret with -Dgcrypt=false even tho libsecret is in GNOME Platform in flatpak too.
>
> And [pasword-manager-service](https://snapcraft.io/docs/password-manager-service-interface) plug seems to be needed anyway.

Just like I thought, it doesn't work in Snap even with `password-manager-service`, libsecret should be built with `-Dgcrypt=false`

PR #579, opened 2023-09-10
Clear issues list (again) for 2023.9.0

### LINKED ISSUE #568: Option to dowload only subtitles ?
Hello, thanks again for your great work with parabole. My suggestion is to have an option to download only subtitles for a movie. The reason is that many times, I have downloaded a movie, and have the choice between SRT or VTT subtitles. But sometimes, the one I choose does not work. Then I have to redownload everything in order to get the subtitles right. It would be much simpler to be able to download only the subtitle file so that I can test which on works. Many thanks again.

### LINKED ISSUE #572: Not all subtitle languages are supported
The function `ValidateSubtitleLangs` returns false for languages that follows this form of scheme: `en-CA`. I believe the issue lies in that the list of codes is a union of ISO letters with either two or three letters. This causes to languages like `en-CA` (Canada) becomes invalid.

https://github.com/NickvisionApps/Parabolic/blob/2ab34a7388b27a5882daaf51f66c50b954eb9825/NickvisionTubeConverter.Shared/Controllers/PreferencesViewController.cs#L29-L30

### LINKED ISSUE #574: Keep original channel name in metadata
One some videos the channel name, that gets written into the metadata, is translated into English, f.i. in [this](https://www.youtube.com/watch?v=akHMCjT9kQA) one, where the original channel name is "Первый образовательный телеканал", but the metadata shows "The first educational TV channel".

Is it possible to define to always use the original channel name, rather the translation?

### LINKED ISSUE #576: Include URL to video in metadata
It would be very convenient, if the URL to the video could be written in to the metadata. F.i. [Video Downloader](https://flathub.org/apps/com.github.unrud.VideoDownloader) already does that and puts it in the commentary field.

fsobolev:
- Closes #568
  No, the option to download only subtitles wasn't added, but instead now there's no format selection, it's chosen by the app to ensure subtitle will be embedded (SRT for MP4 and VTT for WEBM). The state of "Download Subtitle" switch is now also saved to help users who always download video with subtitles.
- Closes #572
  Two-letter codes with region are now supported. Can be tested with [this video](https://www.youtube.com/watch?v=Y-rmzh0PI3c) and `pt-BR` language code.
  Also I removed apply button in the entry, it's a bit confusing and doesn't really serve any purpose: we will save configuration on window closing anyway, no need to save it at some other point.
- Closes #574
  `ffprobe` output for the video linked in the issue:
  ```
  Input #0, matroska,webm, from 'Ваш топинамбур.webm':
  Metadata:
    title           : Ваш топинамбур
    ARTIST          : Первый образовательный телеканал
    track           : 0
    DATE            : 20210615
    ENCODER         : Lavf60.3.100
  ```
  You can see ARTIST is now in Russian.
  This is done by finding current culture's name or two-letter code in a hardcoded list of youtube's supported languages. We have to do this, because giving unsupported language code causes yt-dlp to fail, even with `ignoreerrors`.
- Closes #576
  A switch was added to preferences:
  image
  Off by default. I also noticed there was `handler_name` field in streams' metadata containing info about Google in case of youtube videos, so this is now cleared too.

### DIFF
--- a/NickvisionTubeConverter.Shared/Controllers/PreferencesViewController.cs
+++ b/NickvisionTubeConverter.Shared/Controllers/PreferencesViewController.cs
@@ -219,7 +219,7 @@ public class PreferencesViewController
     }
 
     /// <summary>
-    /// Whether ot not to remove data about media source from metadata
+    /// Whether or not to remove data about media source from metadata
     /// </summary>
     /// <remarks>This includes comment, description, synopsis and purl fields</remarks>
     public bool RemoveSourceData
--- a/NickvisionTubeConverter.Shared/Controllers/PreferencesViewController.cs
+++ b/NickvisionTubeConverter.Shared/Controllers/PreferencesViewController.cs
@@ -26,7 +26,7 @@ public class PreferencesViewController
     {
         var cultures = CultureInfo.GetCultures(CultureTypes.AllCultures).Where(x => !string.IsNullOrEmpty(x.Name)).ToArray();
         var codes = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
-        codes.UnionWith(cultures.Select(x => x.TwoLetterISOLanguageName));
+        codes.UnionWith(cultures.Select(x => x.Name));
         codes.UnionWith(cultures.Select(x => x.ThreeLetterISOLanguageName));
         _supportedLangCodes = codes.ToList();
     }
--- a/NickvisionTubeConverter.Shared/Controllers/PreferencesViewController.cs
+++ b/NickvisionTubeConverter.Shared/Controllers/PreferencesViewController.cs
@@ -219,6 +219,17 @@ public class PreferencesViewController
     }
 
     /// <summary>
+    /// Whether ot not to remove data about media source from metadata
+    /// </summary>
+    /// <remarks>This includes comment, description, synopsis and purl fields</remarks>
+    public bool RemoveSourceData
+    {
+        get => Configuration.Current.RemoveSourceData;
+
+        set => Configuration.Current.RemoveSourceData = value;
+    }
+
+    /// <summary>
     /// Whether or not to embed chapters in a download
     /// </summary>
     public bool EmbedChapters
--- a/NickvisionTubeConverter.Shared/Models/Download.cs
+++ b/NickvisionTubeConverter.Shared/Models/Download.cs
@@ -26,6 +26,8 @@ public enum Quality
 /// </summary>
 public class Download
 {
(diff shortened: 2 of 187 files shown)


nlogozzo:
The preference should be:
Title: `Remove Source Data`
Description: `Clear metadata fields containing the URL and other identifying information of the media source`

NickvisionTubeConverter.Shared/Controllers/PreferencesViewController.cs
```diff
@@ -26,7 +26,7 @@ internal PreferencesViewController()
     {
         var cultures = CultureInfo.GetCultures(CultureTypes.AllCultures).Where(x => !string.IsNullOrEmpty(x.Name)).ToArray();
         var codes = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
-        codes.UnionWith(cultures.Select(x => x.TwoLetterISOLanguageName));
```
nlogozzo, line 29: should this line stay that way it's two letter, two letter with region and three letter?

nlogozzo (changes requested):

NickvisionTubeConverter.Shared/Models/Download.cs
```diff
@@ -21,21 +21,13 @@ public enum Quality
...
-    VTT,
-    SRT
-}
-
 /// <summary>
 /// A model of a media download
 /// </summary>
 public class Download
 {
+    private static string[] _youtubeLangCodes = { "af", "az", "id", "ms", "bs", "ca", "cs", "da", "de", "et", "en-IN", "en-GB", "en", "es", "es-419", "es-US", "eu", "fil", "fr", "fr-CA", "gl", "hr", "zu", "is", "it", "sw", "lv", "lt", "hu", "nl", "no", "uz", "pl", "pt-PT", "pt", "ro", "sq", "sk", "sl", "sr-Latn", "fi", "sv", "vi", "tr", "be", "bg", "ky", "kk", "mk", "mn", "ru", "sr", "uk", "el", "hy", "iw", "ur", "ar", "fa", "ne", "mr", "hi", "as", "bn", "pa", "gu", "or", "ta", "te", "kn", "ml", "si", "th", "lo", "my", "ka", "am", "km", "zh-CN", "zh-TW", "zh-HK", "ja", "ko" };
```
nlogozzo, line 29: I'd make this `internal static string[] YoutubeLangCodes` that why in `MediaUrlInfo` you can do `Download.YoutubeLangCodes` instead of making a second `_youtubeLangCodes`

nlogozzo (changes requested):

NickvisionTubeConverter.Shared/Controllers/PreferencesViewController.cs
fsobolev: `cultures` contains both "general" cultures (with name like `en`) and region-specific ones (`en-US`), so there's no need for 2-letter names.

nlogozzo: so `Names` covers both two letter and two letter with region?

fsobolev: yep

nlogozzo (approved):

PR #601, opened 2023-09-23
Use Aura in DependencyManager + Fix arte support

### LINKED ISSUE #598: It seems that all arte.tv links are now considered as invalid
I have tried to download a video from arte.tv, but Parabole says the link is invalid. I have test others arte.tv link, and get the same invalid message error. However, the links are valid.
Here are some examples:
https://www.arte.tv/fr/videos/114146-000-A/omar-m-a-tuer/
https://www.arte.tv/fr/videos/111195-000-A/face-a-face/
https://www.arte.tv/fr/videos/000430-000-A/charles-mort-ou-vif/

Strange problem ?

soumyaDghosh:
1. Uses the `PATH` environment variable to find dependencies
2. No need of layouts and environment variables in snap
3. Removed the matrix in snap workflow
4. Fix arte.tv support by applying a patch to yt-dlp (Closes #598)

### DIFF
--- a/NickvisionTubeConverter.Shared/Helpers/DependencyManager.cs
+++ b/NickvisionTubeConverter.Shared/Helpers/DependencyManager.cs
@@ -74,19 +74,19 @@ internal static class DependencyManager
     {
         try
         {
-                var process = new Process
+            var process = new Process
+            {
+                StartInfo = new ProcessStartInfo
                 {
-                    StartInfo = new ProcessStartInfo
-                    {
-                        FileName = "python3",
-                        Arguments = "-c \"import sysconfig; import os; print(('/snap/tube-converter/current/gnome-platform' if os.environ.get('SNAP') else '') + ('/'.join(sysconfig.get_config_vars('LIBDIR', 'INSTSONAME'))))\"",
-                        UseShellExecute = false,
-                        RedirectStandardOutput = true
-                    }
-                };
-                process.Start();
-                Runtime.PythonDLL = process.StandardOutput.ReadToEnd().Trim();
-                process.WaitForExit();
+                    FileName = "python3",
+                    Arguments = "-c \"import sysconfig; import os; print(('/snap/tube-converter/current/gnome-platform' if os.environ.get('SNAP') else '') + ('/'.join(sysconfig.get_config_vars('LIBDIR', 'INSTSONAME'))))\"",
+                    UseShellExecute = false,
+                    RedirectStandardOutput = true
+                }
+            };
+            process.Start();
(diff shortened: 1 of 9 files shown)


NickvisionTubeConverter.Shared/Helpers/DependencyManager.cs
```diff
@@ -88,26 +74,19 @@ public static bool SetupDependencies()
...
                         FileName = "python3",
-                        Arguments = "-c \"import sysconfig; print('/'.join(sysconfig.get_config_vars('LIBDIR', 'INSTSONAME')))\"",
+                        Arguments = "-c \"import sysconfig; import os; print(('/snap/tube-converter/current/gnome-platform' if os.environ.get('SNAP') else '') + ('/'.join(sysconfig.get_config_vars('LIBDIR', 'INSTSONAME'))))\"",
                         UseShellExecute = false,
                         RedirectStandardOutput = true
                     }
                 };
                 process.Start();
                 Runtime.PythonDLL = process.StandardOutput.ReadToEnd().Trim();
                 process.WaitForExit();
```
fsobolev, line 89: wrong indentations

fsobolev (changes requested):

fsobolev:
I will fix flatpak later

NickvisionTubeConverter.Shared/Helpers/DependencyManager.cs
```diff
@@ -14,21 +15,16 @@ internal static class DependencyManager
...
             {
-                var path = $"{prefix}/bin/python3";
+                var path = $"{dir}{Path.DirectorySeparatorChar}python3";
                 if (File.Exists(path))
                 {
                     return path;
                 }
             }
-            return "python3";
+            throw new Exception("python3 not in path");
```
nlogozzo, line 27: I'd still `return "python3"` in case we bundle it with the app on windows let's say in the same directory it will just pull it from there. Same for other dependencies.

Or maybe:
```csharp
if(File.Exists("python3'))
{
    return "python3";
}
throw new Exception("python3 not in path");
```
*again same for the other dependencies*

nlogozzo (changes requested):

NickvisionTubeConverter.Shared/Helpers/DependencyManager.cs
fsobolev: I doubt it's good idea to place all dependencies in the same directory. And we could modify PATH to point to all required directories before DependencyManager resolves all deps. So I think for now it's better to keep as it is instead of adding unused code that may be a subject to change.

nlogozzo (approved):

PR #604, opened 2023-09-24
Snap: update yt-dlp to latest tag

soumyaDghosh:


fsobolev (changes requested):
Patch is not needed anymore

DaPigGuy (approved):

fsobolev (approved):

PR #606, opened 2023-09-24
Embed subtitle setting, all subtitles, update yt-dlp in flatpak

### LINKED ISSUE #596: [Feature Request] download all available subtitles

### LINKED ISSUE #599: Parabolic no longer works as it should, after the last update.
Parabolic no longer works as it should, after the last update.
Was using the Snap version, but have switched to the AUR version, but it's the same thing there.
Can't get the subtitles when downloading the video.
During the download, I have to quickly copy the .vtt file to another place, otherwise it will be deleted, in order to get a subtitles file to use.
While troubleshooting, I saw that the operation log file was overwritten during download.
This when I used the "Disallow Conversions - OFF" parameter.
In order to be able to include everything that was done and written in the log file, I had to use a file management program to copy out the log file as it was written to it, so that it would not be overwritten.
When I used the "Disallow Conversions - ON" parameter, the log file was not overwritten, it seems, but it still didn't work as it should.
Sends various log files and other information for troubleshooting.
NOTE. The log files that I send show that the log files are changed during the download of video and subtitles.
I used "Meld" to compare the files.

[Disallow Conversions - OFF - log3-20:16:53.txt](https://github.com/NickvisionApps/Parabolic/files/12676380/Disallow.Conversions.-.OFF.-.log3-20.16.53.txt)
[Disallow Conversions - OFF - log4-20:17:03.txt](https://github.com/NickvisionApps/Parabolic/files/12676381/Disallow.Conversions.-.OFF.-.log4-20.17.03.txt)
[Disallow Conversions - OFF - MediaInfo.txt](https://github.com/NickvisionApps/Parabolic/files/12676382/Disallow.Conversions.-.OFF.-.MediaInfo.txt)
[Disallow Conversions - ON - log6-19:59:56.txt](https://github.com/NickvisionApps/Parabolic/files/12676385/Disallow.Conversions.-.ON.-.log6-19.59.56.txt)
[Disallow Conversions - ON - MediaInfo.txt](https://github.com/NickvisionApps/Parabolic/files/12676386/Disallow.Conversions.-.ON.-.MediaInfo.txt)
[Versions.txt](https://github.com/NickvisionApps/Parabolic/files/12676390/Versions.txt)

### LINKED ISSUE #603: Update yt-dlp in flatpak and snap
Needed to wait just one more day instead of applying patches 😅 
https://github.com/yt-dlp/yt-dlp/releases/tag/2023.09.24

fsobolev:
Closes #596 - if subtitle langs are set to `all` or `_("all")`, all available non-auto-generated subtitles will be downloaded
Closes #599 - custom convertor removes styling from VTT subtitle, allowing ffmpeg to process it correctly
Together with #604 closes #603


nlogozzo:
I have work until 4 PM here, when I get home I'll test and review and then release beta1 :)

nlogozzo:
Maybe we should add a description to the embedded subtitles switching saying  something like "If disabled, downloaded subtitles will be saved to a separate file instead". To signify that the ability to download subtitles isn't effected by the switch.

NickvisionTubeConverter.GNOME/Program.cs
```diff
@@ -35,9 +35,11 @@ public Program(string[] args)
         _mainWindow = null;
         _mainWindowController = new MainWindowController();
         _mainWindowController.AppInfo.Changelog =
-            @"* Added an option to disable subtitle embedding
-              * Fixed an issue where an embedded subtitle was empty
-              * If embedding a subtitle fails, it will now be saved as a separate file
+            @"* Added an option to disable embedding subtitles in a file and instead download them in a separate file
+              * Added the ability to specify ""all"" in the subtitle languages list to download all available languages
+              * Fixed an issue where empty subtitles were sometimes embedded
+              * Fixed an issue where arte.tv links were not validating
```
fsobolev, line 41: That was yt-dlp bug, maybe just say that yt-dlp was updated in Flatpak and Snap fixing some issues?

fsobolev:
> Maybe we should add a description to the embedded subtitles switching saying something like "If disabled, downloaded subtitles will be saved to a separate file instead". To signify that the ability to download subtitles isn't effected by the switch.

What about

`If disabled or if embedding fails, downloaded subtitle will be saved to a separate file instead.`

NickvisionTubeConverter.GNOME/Program.cs
nlogozzo: Since this issue was opened on our app, I mentioned it specifically. 

Maybe `Fixed an issue where arte.tv links were not validating thanks to a yt-dlp update`?

nlogozzo:
> > Maybe we should add a description to the embedded subtitles switching saying something like "If disabled, downloaded subtitles will be saved to a separate file instead". To signify that the ability to download subtitles isn't effected by the switch.
> 
> What about
> 
> `If disabled or if embedding fails, downloaded subtitle will be saved to a separate file instead.`

Perfect!

NickvisionTubeConverter.GNOME/Program.cs
fsobolev: okay 👍🏻

nlogozzo (approved):

PR #610, opened 2023-09-27
SystemDirectories, TaskbarItem, Aura 2023.9.6

fsobolev:
Also added count to taskbar to make it less confusing when progress bar goes backward because number of active downloads has changed.


fsobolev:
@soumyaDghosh libunity isn't needed in snap anymore

NickvisionTubeConverter.Shared/Controllers/MainWindowController.cs
```diff
@@ -21,6 +23,10 @@ public class MainWindowController : IDisposable
     private nint _pythonThreadState;
     private Keyring? _keyring;
     private NetworkMonitor? _netmon;
+    private TaskbarItem? _taskbarItem;
+    private readonly Stopwatch _taskbarStopwatch;
+
+    private const int _taskbarStopwatchThreshold = 500;
```
nlogozzo, line 29: should be named `TASKBAR_STOPWATCH_THRESHOLD` as is the convention for consts

NickvisionTubeConverter.Shared/Controllers/MainWindowController.cs
```diff
@@ -243,6 +251,43 @@ public void UpdateKeyring(KeyringDialogController controller)
         _keyring = controller.Keyring;
     }
 
+    public void SetTaskbarItem(TaskbarItem? taskbarItem)
```
nlogozzo, line 254: Why not make this a property and not nullable:

```csharp
public TaskbarItem TaskbarItem
{
    set
    {
        _taskbarItem = value;
        DownloadManager.DownloadProgressUpdated += (_, _) => UpdateTaskbar();
        DownloadManager.DownloadCompleted += (_, _) => UpdateTaskbar(true);
        DownloadManager.DownloadStopped += (_, _) => UpdateTaskbar(true);
    }
}
```

nlogozzo (changes requested):

nlogozzo:
Could we update to GNOME 45 runtime? The only changes should be to use `Adw.SwitchRow` and `Adw.SpinRow`, correct? All the other widgets and design is fine the way it is...

*Unless I'm wrong then we can save it for later*

NickvisionTubeConverter.Shared/Controllers/MainWindowController.cs
fsobolev: Idk if [these guidelines](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/property?redirectedfrom=MSDN) are relevant for us, but they say:

>  DO NOT provide set-only properties

Also it's nullable because if dbus connection fails TaskbarItem will be null.

nlogozzo: > DO NOT provide set-only properties

Eh, this guideline I'm okay with breaking...we've been doing it and i've seen most PInvoke/Marshal libraries do the same so it;s fine with me.

> Also it's nullable because if dbus connection fails TaskbarItem will be null.

Ah ok so then the header can be `public TaskbarItem? TaskbarItem` and you can add the null check and return inside the set body.

--

But why is this a public method/property anyway? Isn't all the taskbar code, initializing and all, handled in the controller anyway?

fsobolev:
> Could we update to GNOME 45 runtime? The only changes should be to use `Adw.SwitchRow` and `Adw.SpinRow`, correct? All the other widgets and design is fine the way it is...
> 
> _Unless I'm wrong then we can save it for later_

From proposal:

> [ToolbarView](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.ToolbarView.html), we will probably need to use it everywhere. From blog post:
>> It’s recommended to always use it instead of GtkBox when you have header bars or other toolbars, regardless of whether you’re using split views.
> 
> Also note that flat CSS class for headerbars was deprecated.

NickvisionTubeConverter.Shared/Controllers/MainWindowController.cs
fsobolev: > Isn't all the taskbar code, initializing and all, handled in the controller anyway with new Aura?

TaskbarItem creation is platform-specific: `ConnectLinuxAsync` accepts desktop file name, `ConnectWindows` accepts window handle, so TaskbarItem is created in UI and passed to controller.

nlogozzo:
> > Could we update to GNOME 45 runtime? The only changes should be to use `Adw.SwitchRow` and `Adw.SpinRow`, correct? All the other widgets and design is fine the way it is...
> > 
> > _Unless I'm wrong then we can save it for later_
> 
> From proposal:
> 
> > [ToolbarView](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.ToolbarView.html), we will probably need to use it everywhere. From blog post:
> >> It’s recommended to always use it instead of GtkBox when you have header bars or other toolbars, regardless of whether you’re using split views.
> > 
> > Also note that flat CSS class for headerbars was deprecated.

Ah ok...save it for later then 😅

NickvisionTubeConverter.Shared/Controllers/MainWindowController.cs
nlogozzo: > TaskbarItem creation is platform-specific: ConnectLinuxAsync accepts desktop file name, ConnectWindows accepts window handle, so TaskbarItem is created in UI and passed to controller.

Ah ok got it, then this: `Ah ok so then the header can be public TaskbarItem? TaskbarItem and you can add the null check and return inside the set body.` is fine

PR #640, opened 2023-11-04
URL as CLI Arg + WinUI Improvements

### LINKED ISSUE #628: Allow external applications to add downloads
I am working on [Televido](https://github.com/d-k-bo/televido), an application to access media from German public broadcasting services. I want to offer users the ability to download shows. Since some channels need a more advanced downloader such as yt-dlp, I think it would be the best to rely on Parabolic if it is installed. I can image that there are many other projects with similar needs.

---

I propose that Parabolic supports opening URLs via the `Open` method of the `org.freedesktop.Application` interface. Calling this method should start Parabolic and open the “Add Download” dialog with the “Media URL” field prefilled.

Currently, this method call isn't handled and throws an error:

    GDBus.Error:org.freedesktop.DBus.Error.NotSupported: Application does not open files

I think this could be implemented by adding a handler for the [`GApplication::open`](https://docs.gtk.org/gio/signal.Application.open.html) signal and setting the [`G_APPLICATION_HANDLES_OPEN`](https://docs.gtk.org/gio/flags.ApplicationFlags.html#handles_open) flag. I guess that this would also enable the option to supply the URL as a command line argument.

---

For reference, some video players support starting playback like this. E.g. when you have [Celluloid](https://flathub.org/apps/io.github.celluloid_player.Celluloid) installed and open [D-Spy](https://flathub.org/apps/org.gnome.dspy), you can play a video by passing `(['https://youtu.be/8eXj97stbG8'], {})` as a parameter to the `org.freedesktop.Application.Open` call.

nlogozzo:
Closes #628

TODO:
- [x] GNOME - DBus Open Implementation
- [x] WinUI - Install psutil & yt-dlp from installer, not inside app

Better WinUI AddDownloadDialog:

- Validate button in dialog footer itself:
image

- New wait for validation:
image

### DIFF
--- a/NickvisionTubeConverter.GNOME/Program.cs
+++ b/NickvisionTubeConverter.GNOME/Program.cs
@@ -23,7 +23,7 @@ public partial class Program
     /// </summary>
     /// <param name="args">string[]</param>
     /// <returns>Return code from Adw.Application.Run()</returns>
-    public static int Main(string[] args) => new Program(args).Run(args);
+    public static int Main(string[] args) => new Program(args).Run();
 
     /// <summary>
     /// Constructs a Program
@@ -33,9 +33,10 @@ public partial class Program
     {
         _application = Adw.Application.New("org.nickvision.tubeconverter", Gio.ApplicationFlags.FlagsNone);
         _mainWindow = null;
-        _mainWindowController = new MainWindowController();
+        _mainWindowController = new MainWindowController(args);
         _mainWindowController.AppInfo.Changelog =
-            @"* Fixed an issue where aria's max connections per server preference was allowed to be greater than 16
+            @"* A URL can now be passed to Parabolic via the command-line or the freedesktop application open protocol to trigger its validation of startup
+              * Fixed an issue where aria's max connections per server preference was allowed to be greater than 16
               * Fixed an issue where enabling the ""Download Specific Timeframe"" advanced option would cause a crash for certain media downloads
               * Updated translations (Thanks everyone on Weblate!)";
         _application.OnActivate += OnActivate;
@@ -60,17 +61,15 @@ public partial class Program
     /// <summary>
     /// Runs the program
     /// </summary>
-    /// <param name="args">Command-line arguments</param>
     /// <returns>Return code from Adw.Application.Run()</returns>
-    public int Run(string[] args)
+    public int Run()
     {
         try
         {
             SynchronizationContext.SetSynchronizationContext(new GLib.Internal.MainLoopSynchronizationContext());
-            var argc = args.Length + 1;
+            var argc = 1;
             var argv = new string[argc];
             argv[0] = "org.nickvision.tubeconverter";
-            args.CopyTo(argv, 1);
             return _application.Run(argc, argv);
         }
         catch (Exception ex)
--- a/NickvisionTubeConverter.GNOME/Program.cs
+++ b/NickvisionTubeConverter.GNOME/Program.cs
@@ -67,11 +67,7 @@ public partial class Program
     {
         try
         {
-            SynchronizationContext.SetSynchronizationContext(new GLib.Internal.MainLoopSynchronizationContext());
-            var argc = 1;
-            var argv = new string[argc];
-            argv[0] = "org.nickvision.tubeconverter";
-            return _application.Run(argc, argv);
+            return _application.RunWithSynchronizationContext();
         }
         catch (Exception ex)
         {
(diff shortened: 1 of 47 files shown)


DaPigGuy (approved):

NickvisionTubeConverter.GNOME/Program.cs
```diff
@@ -108,4 +108,15 @@ private async void OnActivate(Gio.Application sender, EventArgs e)
...
+
+    /// <summary>
+    /// Occurs when the application is opened
+    /// </summary>
+    /// <param name="sender">Gio.Application</param>
+    /// <param name="e">Gio.Application.OpenSignalArgs</param>
+    private void OnOpen(Gio.Application sender, Gio.Application.OpenSignalArgs e)
+    {
+        _mainWindowController.UrlToLaunch = e.Files[0].GetUri();
+        _application.Activate();
```
fsobolev, line 120: this will call `_mainWindow!.Present()` but will not do anything with UriToLaunch, no?

NickvisionTubeConverter.GNOME/Program.cs
```diff
@@ -60,17 +62,15 @@ public Program(string[] args)
...
     {
         try
         {
             SynchronizationContext.SetSynchronizationContext(new GLib.Internal.MainLoopSynchronizationContext());
-            var argc = args.Length + 1;
+            var argc = 1;
             var argv = new string[argc];
             argv[0] = "org.nickvision.tubeconverter";
-            args.CopyTo(argv, 1);
             return _application.Run(argc, argv);
```
fsobolev, line 74: You could use `_application.RunWithSynchronizationContext` here

fsobolev (changes requested):

NickvisionTubeConverter.GNOME/Program.cs
nlogozzo: The activates the application, which triggers the OnActivate event handler above this method which starts the MainWindow

nlogozzo: RunWithSynchronizationContext doesn't accept args though, no?

Only in 0.5.0 preview it does.

NickvisionTubeConverter.GNOME/Program.cs
nlogozzo: RunWithSynchronizationContext doesn't accept args though, no?

Only in 0.5.0 preview it does.

NickvisionTubeConverter.GNOME/Program.cs
fsobolev:
```csharp
if (_mainWindow != null)
        {
            _mainWindow!.SetVisible(true);
            _mainWindow!.Present();
        }
```

NickvisionTubeConverter.GNOME/Program.cs
nlogozzo: `_application.Run(argc, argv)` passes `org.nickvision.tubeconverter`...I thought that was needed for dbus stuff?

NickvisionTubeConverter.GNOME/Program.cs
nlogozzo: Ah you're right, my bad let me fix that

NickvisionTubeConverter.GNOME/Program.cs
fsobolev: I don't think so. ID is set in `Adw.Application.New`

nlogozzo: Ah ok. So then why did we ever pass them in the first place 😅

fsobolev: maybe it was copied from Denaro or Tagger? 😄

nlogozzo: Ah makes sense...something from the early days 😂 we've come a long way in our apps that sometimes some parts feel like legacy code that should remain untouched hahaha

NickvisionTubeConverter.GNOME/Program.cs
nlogozzo: All fixed in the latest commit.

fsobolev (approved):

nlogozzo:
@fsobolev Quick change for WinUI:

Adding the Disclaimer dialog to show on startup since there is no place for it in About dialog
image

PR #732, opened 2024-02-02
Add zh_CN translate

madoka773:


nlogozzo:
Even if you aren't translating the docs module, could you please add this language to NickvisionTubeConverter.Shared/Docs/po/LINGUAS and create a po file there as well...

madoka773:
done

nlogozzo (approved):

nlogozzo:
Thanks!

nlogozzo:
Question...Weblate is complaining that zh_CN and zh_Hans are both Chinese simplified. Would you be able to merge these translations into one or remove zh_Hans if it is inadequate?

madoka773:
zh_CN is the correct Chinese Simplified language code.
So I think zh_ Hans need to be deleted.

```
madoka@fedora:~$ echo $LANG
zh_CN.UTF-8
madoka@fedora:~$ 
```

If zh_ Hans, the only way I can see the translation is through this

`flatpak run --env="LC_ALL=zh_Hans.UTF-8" org.nickvision.tubeconverter`

nlogozzo:
Ok I'll remove it. Thanks

PR #747, opened 2024-03-04
don't upload the snap to store if a pull request

soumyaDghosh:


nlogozzo:
Can you add this to the rest?

PR #765, opened 2024-04-18
Use "numeric" style class for download progress

mcclurgm:
This is a small cosmetic change. Currently, the download progress label uses proportional figures. This makes it jump around as the numbers are updated. This PR adds the Adwaita style class "numeric" to use tabular numbers that have uniform width.

I have not been successful building the app on my machine, so I have not tested it in the context of the full app. However, I implemented this change to the Blueprint file in Workbench and it had the desired result.


nlogozzo (approved):

nlogozzo:
https://github.com/NickvisionApps/Parabolic/releases/tag/2024.5.0

PR #865, opened 2024-10-06
Add and use display name variable

bragefuglseth:
Parabolic is currently listed in all-lowercase as «parabolic» in app stores, which is not a particularly flattering look.

Skjermbilde fra 2024-10-06 13-08-22

This PR adds a separate variable containing a «pretty» display version of the name. I'm not well-versed in CMake at all, but I suppose there isn't more that needs to be done for it to work.


nlogozzo:
The changes you made are perfect. 

Just want to verify if we can use this new variable in other files.

Maybe for example in the .desktop file?

Could you take a look?

nlogozzo:
I took care of it. Thanks again for the PR :)

PR #953, opened 2024-11-15
Fix building on macOS with libsecret

barracuda156:
Verified to fix the build error both on Sonoma and on 10.6. Otherwise CMake fails to find `libsecret`.

P. S. How does it work on Linux though?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,6 +20,10 @@ set(SHORT_NAME "parabolic")
 set(DISPLAY_NAME "Parabolic")
 include(GNUInstallDirs)
 
+if(APPLE)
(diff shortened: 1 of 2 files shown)


nlogozzo:
This isn't the fix needed.

We forgot to update the config file in the cmake folder in the libnick repo with the new dependency structure...could you take a look at that? 

I am working again all day today and probably won't get to be on my PC.

nlogozzo:
This file needs to be modified with the changes you made in the CMakeLists file: https://github.com/NickvisionApps/libnick/blob/main/cmake/config.cmake.in

barracuda156:
Thanks, I will try that today and rebase the PR then.

nlogozzo:
You can apply it to this same PR: https://github.com/NickvisionApps/libnick/pull/51

PR #1050, opened 2025-01-27
fix(gnome): No ellipses on Video Format combobox

### LINKED ISSUE #1047: GNOME: AddDownloadDialog: Video formats list is truncated
Issue: The combobox with the video formats is truncated with ellipses so it's hard to read what each option is.

Image

```org.nickvision.tubeconverter
Linux
2025.1.2

Deployment Mode: Flatpak
Locale: en_US.utf8
Running From: "/app/lib/org.nickvision.tubeconverter"

yt-dlp version 2025.01.15
ffmpeg version 7.1 
aria2 version 1.37.0

GTK 4.16.12
libadwaita 1.6.3

LANG=en_US.UTF-8
LC_CTYPE="en_US.utf8"
LC_NUMERIC="en_US.utf8"
LC_TIME="en_US.utf8"
LC_COLLATE="en_US.utf8"
LC_MONETARY="en_US.utf8"
LC_MESSAGES="en_US.utf8"
LC_PAPER="en_US.utf8"
LC_NAME="en_US.utf8"
LC_ADDRESS="en_US.utf8"
LC_TELEPHONE="en_US.utf8"
LC_MEASUREMENT="en_US.utf8"
LC_IDENTIFICATION="en_US.utf8"
LC_ALL=en_US.utf8
```

largestgithubuseronearth:
Closes #1047 

Summary:
- Add `GtkHelpers::newNoEllipsesItemFactory()` which creates a `GtkListItemFactory` that disables ellipses on every widget in the ListView.
- This factory has two signal callbacks in `helpers/listitemfactory.cpp` which do the work of creating the label for each ListItem.
- Call the above newFactory function inside `GtkHelpers::setComboRowModel()` and pass the factory into `adw_combo_row_set_factory()`

This is the result in the GUI
Screenshot From 2025-01-26 18-40-29

This works as expected but is a sloppy C solution for a C++ app structure. There's a few things that need to be fixed:
- [x] Applies to all ComboRows indiscriminately. Factory needs to reference the GtkStringList to apply the right string to the ListItem label but the GtkStringList only exists inside `setComboRowModel()`.
- [x]  I don't know where to put the factory callbacks since they aren't part of a class. They should be private but they don't feel right in `GtkHelpers`.
- [x] The factory constructor has a very narrow use case and an awkwardly specific name. It could be reworked to be more generic and take multiple types of attributes or take closures but that might not be worth the effort if you don't need a ListItemFactory anywhere else.

I'm not very familiar with C++ design so I'd appreciate your opinion on how to reconcile these so it fits neatly into the existing code.


nlogozzo:
First off, thank you for this PR! I had no idea this was possible, so thank you again for the implementation!

I committed a more C++ implementation of what you had:
- The free functions in `listitemfactory.cpp` were removed and placed inline of `g_signal_connect` as C++ lambdas 
- Added an `allowElipse` parameter to `setComboRowModel()`, `true` by default, that when set to `false` will trigger the construction of the factory and the prevention of ellipses
- Made the label constructed in the factory setup left aligned instead of center aligned

Let me know if you have any questions on the implementation I committed, I am more than happy to go into more detail!

Thanks again!

PR #1664, opened 2026-02-13
Fix and improve `chmod` in "YtdlpExecutableService.cs" for better permissions

kris3713:
It was previously returning a "missing operand" error. This ensures that `chmod` actually gives executable permissions to `yt-dlp` after downloading it.

I also tested this patch on my end while I was fixing the Fedora package (that I maintain) for this project. It works without any errors.

You can try it out for yourself, with this patch command. I don't know your dev environment, so feel free to replace "path/to" with an actual filepath.

```sh
patch -p1 -s --fuzz=0 --no-backup-if-mismatch -f < path/to/fix_chmod_cmd.patch
```

[fix_chmod_cmd.patch](https://github.com/user-attachments/files/25280359/fix_chmod_cmd.patch)

### DIFF
--- a/Nickvision.Parabolic.Shared/Services/YtdlpExecutableService.cs
+++ b/Nickvision.Parabolic.Shared/Services/YtdlpExecutableService.cs
@@ -97,13 +97,8 @@ public class YtdlpExecutableService : IYtdlpExecutableService
             {
                 using var process = new Process()
                 {
-                    StartInfo = new ProcessStartInfo
+                    StartInfo = new ProcessStartInfo("chmod", ["0755", path])
                     {
-                        FileName = "chmod",
-                        ArgumentList = [
-                          "0755",
-                          path
-                        ],
                         UseShellExecute = false,
                         CreateNoWindow = true,
                     }


nlogozzo:
Thanks for this! I'm going to move it to the 2026.2.3 branch so it gets bundled all together for that update

kris3713:
You're welcome. And thank you for making such a wonderful front-end to `yt-dlp`.

nlogozzo:
@kris3713 - There's a compile time error:
```
Error: D:\a\Parabolic\Parabolic\Nickvision.Parabolic.Shared\Services\YtdlpExecutableService.cs(103,25): error CS0200: Property or indexer 'ProcessStartInfo.ArgumentList' cannot be assigned to -- it is read only
```

I think you need to specify the list in the constructor how I did: https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.-ctor?view=net-10.0#system-diagnostics-processstartinfo-ctor(system-string-system-collections-generic-ienumerable((system-string)))

----

Which then what's the difference between `+x` and `0755`?

kris3713:
@nlogozzo I think I should revert to using curly braces instead of square braces since that didn't cause any compile time errors.

Is it okay with you if I do this?

kris3713:
> Which then what's the difference between `+x` and `0755`?

The difference between those two is that with `+x` you are giving that file just executable permissions. Whereas with `0755`, it also makes the file executable as well as making it read-only to other users and groups in the UNIX file system (i.e. macOS, Linux, FreeBSD etc).

nlogozzo:
> @nlogozzo I think I should revert to using curly braces instead of square braces since that didn't cause any compile time errors.

That's not the fix. `ArgumentsList` is read-only. So you can't use `=` to set it to anything. You have to use the constructor as I did originally. You should make the code:
```csharp
StartInfo = new ProcessStartInfo("chmod", ["0755", path])
{
    UseShellExecute = false,
    CreateNoWindow = true
}
```

PR #1820, opened 2026-04-20
Fix broken homepage link

archisman-panigrahi:
Homepage link on Flatpak page (https://flathub.org/en/apps/org.nickvision.tubeconverter) was broken due to uppercase "Parabolic.html" instead of "parabolic.html"


nlogozzo (approved):

nlogozzo:
Thanks for this!

PR #1912, opened 2026-06-14
fix: Thumbnail crop

Elektronixx:
Fixing thumbnail crop error by changing "%(s)" Conversion to "%(q)" as described in 
https://github.com/NickvisionApps/Parabolic/issues/1910
https://github.com/NickvisionApps/Parabolic/issues/1913
https://github.com/NickvisionApps/Parabolic/issues/1907

This fix has already been tested my myself and does work.

### DIFF
--- a/Nickvision.Parabolic.Shared/Services/YtdlpExecutableService.cs
+++ b/Nickvision.Parabolic.Shared/Services/YtdlpExecutableService.cs
@@ -326,14 +326,14 @@ public class YtdlpExecutableService : DependencyExecutableService, IYtdlpExecuta
                     arguments.Add("--exec");
                     arguments.Add("before_dl:del %(thumbnails.-1.filepath)q");
                     arguments.Add("--exec");
-                    arguments.Add("before_dl:move \"%(thumbnails.-1.filepath)s.tmp.jpg\" %(thumbnails.-1.filepath)q");
+                    arguments.Add("before_dl:move %(thumbnails.-1.filepath)q.tmp.jpg %(thumbnails.-1.filepath)q");
                 }
                 else
                 {
                     arguments.Add("--exec");
                     arguments.Add("before_dl:rm %(thumbnails.-1.filepath)q");
                     arguments.Add("--exec");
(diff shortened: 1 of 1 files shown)


Slitter-the-Dubstep:
i have no plan how to implement your fix and what exec conversions commands are--
is there some way you could help me out to fix my problem as a non-coder?
how do i add this change to my software?

Elektronixx:
> i have no plan how to implement your fix and what exec conversions commands are-- is there some way you could help me out to fix my problem as a non-coder? how do i add this change to my software?

You don't have to.
Unfortunately there isn't an easy way to "add a change to your software" since the program is compiled.
(Think like for example a cake that has been in the oven, that you now cannot "unbake")

There are basically 2 options:
1. Wait for this PR to be approved, the devs to publish a new version and you update your software.
2. Download my fork of the repository (at the top of the page where it says 'Elektronixx:main' then "Code" -> Download ZIP) install 'Windows App SDK' for Windows, open a terminal/cmd in the unpacked target folder and run the command 'dotnet run --project Nickvision.Parabolic.WinUI' 

This would then run the application with the change applied. However wouldn't recommend it for someone who isn't familiar with it, since it isn't that easy.

markazmaxima:
I have the same error message, end up using CMD to download audio.

nlogozzo (approved):

Slitter-the-Dubstep:
Thank you for your efforts and explanation in this issue.
Im hoping that that this fix is going to be implemented into an official release soon.
Sadly your instrutctions of installing your branch didnt work on my system, probably due to user error.

However, if anyone is reading this and is frustrated, ive actually found a valid workaround with tools that id recommend:
**Mp3Tag** is a program that ive already used to edit music-metadata, especially for my library of opus-downloads.
Turns out you can also **losslessly crop thumbnails / album covers in bulk**. Thats the method id recommend for a straightforward post-hoc fix.

Other than that, theres also **MusicBrainz Picard**.
Thats a software that automatically **scans your files and searches their databases for fitting metadata, including albumcover.**
It might not always get all the metadata perfectly right though, so just make sure to check, that it didn't mess anything substantial up.

